Issue #25, opened 2016-07-23
Column header don't move when do horizontal scroll on table body.


wor-k:
Column header don't move when do horizontal scroll on table body. I testing on 0.1.14 version.

amcdnl:
Noticed your PR, does that cover this ?

wor-k:
I send the other PR for this.

amcdnl:
Fixed in 0.2.0

PR #28, opened 2016-07-25
fix Column header on horizontal scroll

wor-k:


amcdnl:
What issue does this code solve?

wor-k:
Its a #25.

amcdnl:
Gotcha; I'm planning on addressing this as part of a larger change.  I'm essentially going to port my code from my ng1 table - https://github.com/swimlane/angular-data-table/blob/master/src/components/header/HeaderController.js#L64

Let me look over this and I'll provide feedback and/or merge :). Thanks!

wor-k:
It's ok. do what you plan. I'm really like this library.

amcdnl:
Thank you 👍 ! I'll keep you posted, I have to get quite a few items done before ngAir on the 16th this will be part of it for sure!

PR #38, opened 2016-07-27
Fix incorrect rxjs imports

DzmitryShylovich:


amcdnl:
Thanks for your contribution!

PR #40, opened 2016-07-28
Provide the ability to specify custom pipe for table rows

DzmitryShylovich:


src/components/DataTable.ts
```diff
@@ -47,131 +50,154 @@ import './datatable.scss';
...
-  ngAfterViewInit() {
+  ngAfterViewInit(): void {
     this.adjustColumns();
 
-    setTimeout(() => {
-      for(let col of this.columns.toArray()) {
-        this.options.columns.push(new TableColumn(col));
-      }
-    });
+    if (this.columns.length) {
```
DzmitryShylovich: Be careful with `setTimeout` because it triggers change detection for the hole app

src/components/body/BodyCell.ts
```diff
@@ -24,22 +24,21 @@ import { TemplateWrapper } from '../../directives/TemplateWrapper';
...
-  directives: [ TemplateWrapper ]
+  directives: [ TemplateWrapper ],
+  changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class DataTableBodyCell {
 
   @Input() column: TableColumn;
   @Input() row: any;
 
+  @HostBinding('class.datatable-body-cell') isBodySell: boolean = true;
```
amcdnl: `isBodySell` typo ;)

src/components/body/BodyRow.ts
```diff
@@ -38,22 +38,19 @@ import { DataTableBodyCell } from './BodyCell';
...
 
-  get isSelected() {
-    return this.state.selected &&
-      this.state.selected.indexOf(this.row) > -1;
+  constructor(private state: StateService) {
   }
 
-  constructor(private state: StateService, elm: ElementRef) {
-    elm.nativeElement.classList.add('datatable-body-row');
+  @HostBinding('class.datatable-body-row') isBodyRow: boolean = true;
```
amcdnl: This is slower since it has to evaluate the binding vs just adding the class.

src/components/DataTable.ts
```diff
@@ -47,131 +50,154 @@ import './datatable.scss';
...
-  @ContentChildren(DataTableColumn) columns;
+  @ContentChildren(DataTableColumn) columns: QueryList<DataTableColumn>;
 
   private element: HTMLElement;
-  private rowDiffer: any;
-  private colDiffer: any;
+  private rowDiffer: IterableDiffer;
+  private colDiffer: IterableDiffer;
+
+  @HostBinding('class.datatable') isDatatable: boolean = true;
```
amcdnl: This is slower than just adding the class due to evaluation.

src/components/DataTable.ts
amcdnl: Gotcha, couldn't get it to work w/o it though. Open to ideas.

src/components/body/BodyCell.ts
DzmitryShylovich: it should be fixed here https://github.com/swimlane/angular2-data-table/pull/40/commits/a286f3683ba76f941a467fa845e538b01b1e87d1

src/components/DataTable.ts
```diff
@@ -47,131 +50,154 @@ import './datatable.scss';
     DataTableFooter,
     Visibility
   ],
-  host: {
```
amcdnl: Why is it more ideal to do it the other way? Just curious, this seems a bit cleaner given the implementation detail.

src/components/DataTable.ts
```diff
@@ -114,13 +114,13 @@ export class DataTable implements OnInit, DoCheck, AfterViewInit {
       let chngd: boolean = false;
       colDiff.forEachAddedItem(() => {
         chngd = true;
-        return;
+        return false;
```
amcdnl: Curious what this broke?

src/components/body/BodyRow.ts
DzmitryShylovich: Yeah, I thought about that. But it's not the "angular" way to directly access native element and add required classes.

src/components/body/Body.ts
```diff
@@ -12,8 +12,7 @@ import { Subscription } from 'rxjs/Subscription';
   selector: 'datatable-body',
   template: `
     <div>
-      <datatable-progress
```
amcdnl: I actually prefer this pattern, similar how React devs do attributes. Any reasons why you did this?

src/components/DataTable.ts
DzmitryShylovich: https://angular.io/styleguide#!#06-03

src/components/body/ProgressBar.ts
```diff
@@ -13,4 +13,8 @@ import { Component, ChangeDetectionStrategy } from '@angular/core';
   `,
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class ProgressBar {}
+export class ProgressBar {
+
+  @Input() visible: boolean = false;
```
amcdnl: If your controlling the visibility on parent, why pass visible here?

src/components/body/BodyRow.ts
amcdnl: Gotcha. I had it like this originally but switched to this given it was faster. Given perf and number of rows its super important to get every inch of perf we can.

I think we should raise this to the ng team for feedback.

src/components/DataTable.ts
amcdnl: LOL, why did they add it if they don't want you to use it. Ok, thanks!

amcdnl:
Once we align on all the items noted, I'll be happy to merge 👍 ! Thanks again!

src/components/body/BodyRow.ts
DzmitryShylovich: I guess this can be pretty easy optimized by the angular compiler in the upcoming rc5. But u're right, would be nice to hear feedback from the core team.

src/components/body/ProgressBar.ts
DzmitryShylovich: Agree, I'll revert these changes.

src/components/body/BodyRow.ts
amcdnl: Lets pull back until we get a clear answer on this, perf is too important for this project for convention over performance.

DzmitryShylovich: ok

DzmitryShylovich:
Fixed.

amcdnl:
Thanks! You didn't have to revert all those style changes, just that add class one.

PR #42, opened 2016-07-31
Created index.d.ts for npm package.

Alber70g:
Created index.d.ts for npm package. This will also fix the dependency resolving in projects using angular-cli


amcdnl:
Haha was it really that easy? #facepalm

PR #47, opened 2016-08-11
chore(build): enable hot module reloading

d3viant0ne:
- Implements Hot Module Reloading via `env.HMR`
- Webpack config is now a function
- Fixes linting errors based on existing config
- Sets `no-console` to warn from error temporarily


amcdnl:
Awesomeness!

PR #49, opened 2016-08-12
(fix): missing px in columnGroupWidths

wor-k:


amcdnl:
Thanks for your contribution!

PR #62, opened 2016-08-16
Add onScroll event for paging

Tempus35:
Added event to body to send when a page was changed on scroll.


amcdnl:
This should be fixed in 0.2.5

PR #67, opened 2016-08-18
Temporarily use 'typings' instead of 'types' for pre-2.0 users.

DanielRosenwasser:
This should help out users who aren't using 2.0 yet as I brought up on https://github.com/swimlane/angular2-data-table/issues/1#issuecomment-240618399.


amcdnl:
Thanks so much @DanielRosenwasser !

PR #80, opened 2016-08-24
Adding datatable-row-odd / even classes to rows

ocombe:
Adding `datatable-row-odd` & `datatable-row-even` classes to rows, based on their $$index


amcdnl:
Thanks!

PR #83, opened 2016-08-25
Allow changing of body height via table options

Tempus35:


amcdnl:
Discussion in Gitter:

> This is a internal property. The property is set at runtime by a calculation of the height of the element. Verify this: https://github.com/swimlane/angular2-data-table/blob/master/src/components/DataTable.ts#L128

amcdnl:
@Tempus35 any updates on this?

Tempus35:
Not yet, sorry been busy. I will look more into it this afternoon or tomorrow though.

Tempus35:
So I did some quick testing, the height of the table is indeed calculated on runtime by amount of rows and row height. Further more the bodyHeight, when set with this modified code, does not change anything. I am closing this.

PR #104, opened 2016-09-08
Modify Datatable to allow multiple tables on a page

Tempus35:
I changed the state service to use maps based on a key that was unique per table. I also changed all the components to use the new system. It passed npm test and was working in npm start. 

I did realize later that the package.json has angular2-data-table.cjs.js instead of angular2-data-table.umd.js and I changed the tsconfig to emit helpers. Feel free to remove both those changes but I needed them for my build.


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "angular2-data-table",
   "version": "0.3.8",
   "description": "angular2-data-table is a Angular2 component for presenting large and complex data.",
-  "main": "release/angular2-data-table.umd.js",
+  "main": "release/angular2-data-table.cjs.js",
```
amcdnl: Why did you change this to CJS vs UMD?

release/components/DataTable.d.ts
```diff
@@ -16,6 +16,7 @@ export declare class DataTable implements OnInit, DoCheck, AfterViewInit {
     private element;
     private rowDiffer;
     private colDiffer;
+    private key;
```
amcdnl: please add a line break after this

amcdnl:
Few things:
- Why was lodash added? IMO lodash is not ever needed anymore.
- I don't use Typings with this project, so those should be removed.
- In the future, don't check in anything inside release. It makes merges hell and harder to review.

After reviewing this, TBH the key solution is really gross. It would be simplier/cleaner just to pass the state container around IMO. I'll take a run at that and we can colloab.

Tempus35:
Lodash was used because I wanted to use some of its functions, yes you can remove them and come up with something else, I just did not know how. As for typings that was only added for lodash and the key solution might not be the best solution but best I could come up with.

package.json
Tempus35: Yea, mentioned that in the initial comment, I have to use cjs for my project, forgot to change back after testing.

amcdnl:
Right on, we can iterate on this. Lodash was needed mainly pre-es5 for Array helpers/etc. 

I've actually never used it and just always used polyfills which since we have core-js you always have them all now.

amcdnl:
Fixed via native option, no longer needed. Thanks for your effort!

PR #107, opened 2016-09-09
Bug fix onPaged event emit wrong property name

zxshinxz:
Updated Pager component onPaged event property name from 'value' => 'page' to match DataTable component event onPageChanged

### DIFF
--- a/src/components/footer/Pager.ts
+++ b/src/components/footer/Pager.ts
@@ -112,7 +112,7 @@ export class DataTablePager {
 
       this.onPaged.emit({
         type: 'pager-event',
-        value: page
+        page: page
       });
     }
   }


src/components/footer/Pager.ts
```diff
@@ -112,7 +112,7 @@ export class DataTablePager {
 
       this.onPaged.emit({
         type: 'pager-event',
-        value: page
+        page: page
```
amcdnl: We should use `value` instead as there are other places that are looking for that.

Tempus35: Need to change this then
` onPageChanged(action) {
    this.state.setPage(this.key, action);
    this.onPageChange.emit(action.page);
  }
`
since it is receiving the event and has no page property.

amcdnl:
Resolved in latest. Thanks!

PR #112, opened 2016-09-12
fixed dependencies for rc6

ocombe:
I removed the rxjs & zonejs peer dependencies, as @angular already requires them for 2 reasons:
- angular 2 is loose on the zone js dependency, which mean that we can use a newer version when needed (for example 0.6.17 has a serious bug with longtrace)
- angular 2 requires rxjs beta 11 when your lib wanted beta 6


amcdnl:
Thanks!

PR #118, opened 2016-09-13
Bug Fix #84 last pagination page shows wrong items

### LINKED ISSUE #84: last pagination page shows wrong items
Hi all,

Thanks for this great library!
I just found a small issue, if the last page is called, in some cases there are still 'old' items on it.
https://plnkr.co/edit/1hIoLSJ8TubcX2pow9Ga?p=preview

If you jump to the last page you see first 7 items, click now on page 12 -> 5 items, again -> 3 items
image

image

zxshinxz:
If DataTable is set to use pagination, removed Body's rows cache.

Added externalPaging condition because I do not know the impact of removing Body's rows cache
on vertical scroll.

### DIFF
--- a/src/components/body/Body.ts
+++ b/src/components/body/Body.ts
@@ -141,8 +141,12 @@ export class DataTableBody implements OnInit, OnDestroy {
     let idx = 0;
     let rowIndex = idxs.first;
 
-    let endSpliceIdx = refresh ? this.state.rowCount : idxs.last - idxs.first;
-    this.rows.splice(0, endSpliceIdx);
+    if (this.state.options.externalPaging) {
+      this.rows.splice(0, this.rows.length);
+    } else {
+      let endSpliceIdx = refresh ? this.state.rowCount : idxs.last - idxs.first;
+      this.rows.splice(0, endSpliceIdx);
+    }
 
     while (rowIndex < idxs.last && rowIndex < this.state.rowCount) {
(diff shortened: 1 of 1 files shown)


src/components/body/Body.ts
```diff
@@ -141,8 +141,12 @@ export class DataTableBody implements OnInit, OnDestroy {
     let idx = 0;
     let rowIndex = idxs.first;
 
-    let endSpliceIdx = refresh ? this.state.rowCount : idxs.last - idxs.first;
-    this.rows.splice(0, endSpliceIdx);
+    if (this.state.options.externalPaging) {
```
amcdnl: Is this really only isolated to external paging?

amcdnl:
I was looking at my past project for ng1 that implements a similar pattern, this might be helpful reference: https://github.com/swimlane/angular-data-table/blob/master/src/components/body/BodyController.js#L324

zxshinxz:
[reference](https://github.com/swimlane/angular-data-table/blob/master/src/components/body/BodyController.js#L358) eventually removes all the rows before attaching new ones. I guess it's safe remove all the rows? 

`updateRows(refresh?: boolean)`

But, then refresh arguments becomes useless.
I actually have not found any usage of updateRows(true) in the source.

amcdnl:
@zxshinxz - was this fixed along with #121?

amcdnl:
Closing due to inactivity.

PR #122, opened 2016-09-14
Fix #121 Only display filtered data

### LINKED ISSUE #121: Displaying wrong items after filtering
### Current behaviour

When filtering data updateRows method is not removing expected rows as shown below : 

Not filtered data : 
capture d ecran 2016-09-14 a 10 42 48

Filtered data : 
capture d ecran 2016-09-14 a 10 46 21
Row count is fine but not displayed data.

### Expected behaviour

Filtered data should display rows starting with 'u' as below : 
capture d ecran 2016-09-14 a 10 55 43

### Reproduction of the problem

Any filtered table because wrong calculation : 
capture d ecran 2016-09-14 a 11 12 11
endSpliceIdx should be 5 but actual is 1. Using slice instead of splice should be ok too.

I think this issue is somehow related to #84 and #118

jntakpe:
Fixes #121

### DIFF
--- a/src/components/body/Body.ts
+++ b/src/components/body/Body.ts
@@ -142,7 +142,7 @@ export class DataTableBody implements OnInit, OnDestroy {
     let rowIndex = idxs.first;
 
     let endSpliceIdx = refresh ? this.state.rowCount : idxs.last - idxs.first;
-    this.rows.splice(0, endSpliceIdx);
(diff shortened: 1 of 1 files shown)


amcdnl:
Awesome! Thanks for your contribution too!

jntakpe:
You're welcome 👍

PR #129, opened 2016-09-15
Fix horizontal header issue

Tempus35:
Adds ability to have horizontal scroll without vertical scroll and have the header reflect the scroll.


amcdnl:
Doh! Thanks man!

PR #142, opened 2016-09-24
Allow configuration table Height in pixels with TableOptions property tableHeight.

amiller29au:
I wanted a way to enable the table height to be configurable and could not find an option to do this.

I using the table option with scrolling, not pagination.

### DIFF
--- a/src/demos/virtual.ts
+++ b/src/demos/virtual.ts
@@ -44,7 +44,8 @@ export class App {
     headerHeight: 50,
     footerHeight: 50,
     rowHeight: 50,
-    scrollbarV: true
+    scrollbarV: true,
+    tableHeight: 500
   });
 
   constructor() {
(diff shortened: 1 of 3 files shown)


amcdnl:
Cool, thanks for the contribution!

amcdnl:
So I'm looking back at this, why would you not set the element's height and use scrollbarV?

amiller29au:
@amcdnl - I am using scrollbarV option only all my grids. I tried setting the element's height but because the bodyHeight was set to 300, it would only render 5 rows, it didn't matter what I had tried.

amcdnl:
See if it works now in 0.7.4. I think it was a bug I fixed last night

amiller29au:
Sorry this did not work, I changed the height of the table using css and number of rows displayed to not increase.

amcdnl:
Can u make a demo of what u are seeing? I can't seem to repo this.

amiller29au:
Ok all good this now working. You could remove redundant code.

PR #143, opened 2016-09-24
Fix issue when specifying column width in HTML Element.

amiller29au:
When calculating column widths they are not converted to number for the width calculation. 

The width is a string value and the width calculation is concat strings rather than summing column widths.

### DIFF
--- a/src/utils/math.ts
+++ b/src/utils/math.ts
@@ -10,7 +10,7 @@ export function columnTotalWidth(columns: any, prop?: any) {
 
   for(let column of columns) {
     const has = prop && column[prop];
-    totalWidth = totalWidth + (has ? +column[prop] : +column.width);
+    totalWidth = totalWidth + (has ? column[prop] : column.width);
   }
 
   return totalWidth;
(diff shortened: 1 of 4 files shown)


src/utils/math.ts
```diff
@@ -122,31 +122,31 @@ export function forceFillColumnWidths(allColumns: any, expectedWidth: any, start
...
-        column.$$oldWidth = column.width;
+        column.$$oldWidth = +column.width;
       }
 
       const newSize = column.$$oldWidth + additionWidthPerColumn;
-      if(column.minWith && newSize < column.minWidth) {
-        column.width = column.minWidth;
-      } else if(column.maxWidth && newSize > column.maxWidth) {
-        column.width = column.maxWidth;
+      if(column.minWith && newSize < +column.minWidth) {
```
amcdnl: Rather than do the coercion several times, lets do it once in the top of the function and just use those variables.

Or maybe even better and use a getter in the column object to do this automatically.

amiller29au:
I agree that using a getter in the column object would be a better place to do this.

We could just convert the value for a number inside the setter, this way conversion is taken place at the source.

amcdnl:
Cool, wanna update that and we can merge?

amiller29au:
Will do tonight

amiller29au:
Wasn't sure how to handle maxWidth so I left it as is.

amcdnl:
Can you rebase? I just made some file name changes :S

PR #144, opened 2016-09-24
Fix issue when column height when column value is null

amiller29au:
The column height is not correct when column value is null.

Main issue here is specifying column borders in stylesheet when value is null, the column border is only rendered for the top part of the column not the full height.

### DIFF
--- a/src/components/body/BodyCell.ts
+++ b/src/components/body/BodyCell.ts
@@ -37,11 +37,11 @@ export class DataTableBodyCell {
     return userPipe ? userPipe.transform(prop) : prop;
   }
 
-  @HostBinding('style.width') get width() {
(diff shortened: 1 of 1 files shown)


amcdnl:
We probably need to add this- https://github.com/swimlane/angular2-data-table/blob/master/src/components/body/BodyCell.ts#L40 and add a check for if its Nan too not to append the px.

amiller29au:
I agree need to add the px suffix or we could simple do @HostBinding('style.width.px') without the need to specify the units in the method.

Not sure how the check if Nan would not append the height to the host element.

amcdnl:
If you wanna update just to do that manually for now, I will merge.

amiller29au:
Will do

PR #145, opened 2016-09-24
Add Row Double Click event handler

amiller29au:
Add ability to capture dblclick event handler on a row using existing onRowClick event.

As can investigate MouseEvent.type value of "dblclick" in application.


amcdnl:
So the double click would do the same as click? I think we should make this an option too, example: to select I could use single or double click. Thoughts?

amiller29au:
Interesting thought!

I will be making so additional improvements today around the selection process, as I have a requirement to  only to allow selection of certain rows based on a defined criteria. My initial thought is is the rowClick event could return a cancel option and if this is set then do not call selectRow().

This way depending on the use case single or double click could allow/disallow a selection.

amiller29au:
It would be nice to specify an interface for the event args being passed to the onRowClick emit method.

What are your thoughts?

src/components/body/Body.ts
```diff
@@ -180,8 +180,12 @@ export class DataTableBody implements OnInit, OnDestroy {
   }
 
   rowClicked(event, index, row): void {
-    this.onRowClick.emit({event, row});
-    this.selectRow(event, index, row);
+    let options = { cancel: false };
```
amcdnl: You are resetting the options object in the listener of row click? Thats interesting but I'm scared of race cases with that.

amcdnl:
Ya, changing `onRowClick` to emit `{ type: single|double, value: row }` like I did [https://github.com/swimlane/angular2-data-table/blob/master/src/components/footer/Pager.ts#L115](here) would be nice.

src/components/body/Body.ts
amiller29au: Inside the rowClick event handler, I'm changing the options property of cancel to true if the condition is met.

I don't expect a race condition to occur.

amiller29au:
That sounds like a great addition. I still think it still needs to pass the MouseEvent as well.

amiller29au:
That sounds like a great addition. I still think it still needs to pass the MouseEvent as well.

amcdnl:
@amiller29au Sure. Whats the use case for mouse event? pass it as `{ event: theevent }`

amiller29au:
The use case could be possibly add menu on right mouse button click. 

So yes pass it as { event: theevent }

amcdnl:
Ya, thats true. Wanna add those in this PR too?

amiller29au:
Closed PR and resubmitted.

PR #146, opened 2016-09-26
Method guard State.setRows

amiller29au:
When consuming the table, I set rows to null when there are no rows present, so need the rows to be cleared. This is especially the case when the rows have not been initialized.

This causes a flow on affect in DataTableBody.ngOnInit() where reading the rows causes an exception.

### DIFF
--- a/src/services/State.ts
+++ b/src/services/State.ts
@@ -82,10 +82,8 @@ export class StateService {
   }
 
   setRows(rows: Array<any>): StateService {
-    if (rows) {
-      this.rows = [...rows];
-      this.onRowsUpdate.emit(rows);
-    }
+    this.rows = rows === undefined || rows === null ? [] : [...rows];
+    this.onRowsUpdate.emit(rows);
     return this;
   }
 


amcdnl:
Hmmm, by setting them to null wouldn't that de-reference the array? In the examples, I specifically make them not splice into the current one to prevent this.

I'm ok with the change to have this ability, just wanna make sure we don't break anything.

jtomaszewski:
IMHO this way is more valid - it should always reset the reference of the array when we call setRows(). It will help later, if we want to change angular2-data-table to not mutate the data. (currently it mutates it - by f.e. adding $index attributes to the each row's object)

If we want to remember the each row's value, we should use trackBy or something like that.

amcdnl:
What do you mean ` Right now it's being deferenced anyway`? I'm not doing that anywhere.

Ya, it works fine if you put an empty array.

amiller29au:
@amcdnl @jtomaszewski - I certainly could put the datatable inside an *ngIf and not display the table, the issue is that a null or undefined could set the rows property and the rows property is not initialized see [https://github.com/swimlane/angular2-data-table/blob/master/src/services/State.ts#L16](url) hence if rows is not initialized to an empty array then this throw an exception [https://github.com/swimlane/angular2-data-table/blob/master/src/components/body/Body.ts#L100](url)

jtomaszewski:
Hey @amiller29au , can you tell if it has been fixed by https://github.com/swimlane/angular2-data-table/commit/d10dcd133eeb75aa22a012570557296101224b08 ?

amiller29au:
It possibly will, but what action should be taken if for whatever reason someone passes null or undefined?

jtomaszewski:
IMHO it should raise an error (that's an invalid value).

Currently it does nothing.

It could as well set the rows to empty array.

amiller29au:
@jtomaszewski , @amcdnl - what is the outcome of method guarding StateService.setRows() ? 

Should the rows be reset to an empty array, rather than the value ignored?

amcdnl:
I'm hoping to address this in a rewrite of inputs to make it reactive. I'm going to close this for now.

PR #151, opened 2016-09-28
In DataTable component, set rows, options & selected inputs in ngOnChanges, instead of ngOnInit

### LINKED ISSUE #149: Performance optimizations
Is the angular2-data-table optimized when it comes to performance?

I didn't get to see through all the code, but after a quick glimpse:
- we could add a `trackBy` option to `*ngFor="let row of rows; let i = index;"` in `Body` component's template
- components like `BodyCell`, `BodyRow` should propably be dependant only on Inputs and use `ChangeDetectionStrategy.OnPush`
- components like `Body` propably could also use `ChangeDetectionStrategy.OnPush`, but as it's currently heavily dependant on `StateService`, we would need to add some rxjs observables to that or something.

What do you guys think? Maybe I could help a bit with that.

jtomaszewski:
Resolves #149 and propably #146 as well.


amcdnl:
Looks good. In the future, please don't run the build. It causes merge conflicts.

PR #152, opened 2016-09-28
Use ChangeDetectionStrategy.OnPush in BodyCell component

jtomaszewski:


amcdnl:
Need to rebase :(.

Does it work with the previous issue I reverted it to?

jtomaszewski:
Yeah, it should work with the previous issue.

jtomaszewski:
Although right now you made bodycell dependent on state, so we'd need to refactor the whole thing a bit to accomplish that ( https://github.com/swimlane/angular2-data-table/pull/144/commits/d16833172189d4fec4a18f959b9bbb5407775f96#diff-6c5fe72c2cab8659bc1ddbfa686ed2f9R29 ).

amcdnl:
I think we could set that property via the parent or pass it down and it would be fine.

amcdnl:
Closing this given comments above.

PR #154, opened 2016-09-28
Refactor selection: try not to mutate the selection array if possible

jtomaszewski:
- add `mutateSelectionState: boolean` to TableOptions (keep true as a default value, to support old behaviour)
- add `rowIdentityFunction: function` to TableOptions


jtomaszewski:
I tested it in the selection demo and my own app, seems to work correctly. The `rowIdentityFunction` can be used later on in our `trackBy` functions, which I should implement soonish.

amcdnl:
Since you built the release, I will have to merge this manually since I included your other PR.

Can you explain those new methods at a high level? Also, should probably add those notes to the docs.

jtomaszewski:
Sorry. Maybe we can write in README contributing section that PRs shouldnt include release/ files.

I wrote some docs in the TableOptions.ts file. Let me know if is it enoguh? Propably we'd need some examples. I could add them into the docs.

src/services/State.ts
```diff
@@ -76,24 +77,21 @@ export class StateService {
...
 
-  setSelected(selected: any[]): StateService {
-    if (!this.selected) {
-      this.selected = selected || [];
-    } else {
-      this.selected.splice(0, this.selected.length);
-      this.selected.push(...selected);
-    }
+  get selected(): any[] {
+    return this.rows.filter(row => this.isRowSelected(row));
```
amiller29au: Could this be possibly slow if there a lots of rows?

amcdnl: Ya, I'm not sure a better way :(

amiller29au: Could `get `selected()` and `setSelected(selected: any[])` be merged as the get/set property

jtomaszewski: Well, in case we'd have like 100.000 rows, yeah, that could be slow. But that's not a problem, we could refactor setSelected and setRows to cache the `selected` value. I'll remake it.

`setSelected` - I'd leave it as it is, just like `setRows` and `setOptions`. Personally, I use getters/setters only if they are very simple (set/get value), but don't have any extra behaviour (like emitting events or sth).

jtomaszewski: Done in https://github.com/swimlane/angular2-data-table/pull/154/commits/3fb87f80042a23ac3f386ed6480fe7827494263a .

amiller29au:
I'm thinking that it would be better to have a new class that its sole purpose to to handle keeping track of row selection. 

In order to achieve this DataTable class has an @Input property that accepts the the selection class. If non is provide a default selection class is created internal. This way if someone wants to override functionality just create a descendant or hook into the events of the selection and changed it to suit use case.

Thoughts?

This would handle the use case in pull request #145 except for the dblclick option.

java2kus:
@amiller29au I agree with you that we need a separate class for handling of the row selections.   Also, we have a use case where clicking on the row for selection is not something that is desirable.   We have links in the table cells which need to trigger a popover menu.   

One other thing I observed is that having a separate selection array to track row selections may not work out if the use case requires that the row selections be preserved on sort.   I am working on a separate PR for detail rows  where I am planning to put the expandable state inside the row object just like the $$index property.

@jtomaszewski @amcdnl @amiller29au what do you think of the above?

jtomaszewski:
> One other thing I observed is that having a separate selection array to track row selections may not work out (unless we have a rowIdentity function where the identity will be preserved and unique) if the use case requires that the row selections be preserved on sort. I am working on a separate PR for detail rows where I am planning to put the expandable state inside the row object just like the $$index property.

If I understand you correctly, I've done it already here: https://github.com/swimlane/angular2-data-table/pull/154/commits/c8162ded254598c7296317cf8785e7dacaeef465 . Look at the `rowIdentityFunction` and `selectedIdentities` array. They are preserved when f.e. you change current rows by changing current page, and then go back - they are still selected.

amcdnl:
You don't need a `rowIdentityFunction`, there is a property called `$$index` bound to every row. https://github.com/swimlane/angular2-data-table/blob/master/src/components/body/body.component.ts#L164

java2kus:
@amcdnl - The `$$index` changes whenever you do a sort or filter on the rows array.   If we link the selection array to the index and not the row then the comparison would fail in the absence of a unique row identity which is maintained across sorts and filters.   `$$index` is a running number for the currently active rows in the stateservice.  Isn't that the case?   This can be resolved by having a running number in `$$identity` which is retained across filter/sorts and provide a custom `rowIdentity` function for those who desire to set their own identity (most would have a unique identity in their dataset).

Please check the below piece of code.

https://github.com/swimlane/angular2-data-table/blob/master/src/components/body/body.component.ts#L188-L191

amcdnl:
@java2kus - Snap! I didn't even think about that. Good point.

src/components/footer/pager.component.ts
```diff
@@ -49,7 +50,8 @@ import {
         </a>
       </li>
     </ul>
-  `
+  `,
+  changeDetection: ChangeDetectionStrategy.OnPush
```
amcdnl: This is there now :)

amcdnl:
I manually merged the majority of this. I'm going to rewrite the selection and rows to be reactive so hopefully that will address the rest of this.

Issue #127, opened 2016-09-15
Visual Glitch when Rows array is static.


RicardoVaranda:
Steps to reproduce:

Set rows as:

```

<datatable
  class='material'
  [rows]='rows'
  [options]='options'
  (onRowClick)="userClicked($event)">
  <datatable-column name="Name" [width]="300"></datatable-column>
  <datatable-column name="Gender"></datatable-column>
  <datatable-column name="Age"></datatable-column>
  <datatable-column name="City" [width]="300" prop="address.city"></datatable-column>
  <datatable-column name="State" [width]="300" prop="address.state"></datatable-column>
</datatable>


    {
      "id": 0,
      "name": "Ramsey Cummings",
      "gender": "male",
      "age": 52,
      "address": {
        "state": "South Carolina",
        "city": "Glendale"
      }
    },
    {
      "id": 1,
      "name": "Stefanie Huff",
      "gender": "female",
      "age": 70,
      "address": {
        "state": "Arizona",
        "city": "Beaverdale"
      }
    },
    {
      "id": 2,
      "name": "Mabel David",
      "gender": "female",
      "age": 52,
      "address": {
        "state": "New Mexico",
        "city": "Grazierville"
      }
    }

  options = new TableOptions({
    columnMode: ColumnMode.force,
    headerHeight: 50,
    footerHeight: 0,
    rowHeight: 50,
    scrollbarV: true,
    scrollbarH: true
  });
```

Result:

![image](https://cloud.githubusercontent.com/assets/6951312/18554068/f13bca40-7b5a-11e6-97eb-cbdcbbd58bbe.png)

After the scroll bar is pressed it does update to the intended tho:

![image](https://cloud.githubusercontent.com/assets/6951312/18554094/095e02dc-7b5b-11e6-9a80-a2d3d94dc7d4.png)

I have used Augury and seen that the the array does not initially have the $$index set to anything only after I click the scroll bar does it update the rows and appends to each object.

Disabling scrollbarV fixes this issue so I think the problem is with the functionality that variable has.

amcdnl:
Please create a demo? :)

RicardoVaranda:
Can you provide me with a jfiddle template for the project?

MonicaOlejniczak:
I've just stumbled upon this same issue. Here is a simple example demonstrating the issue based on the Horizontal/Vertical Scrolling example found in the documentation.

[...]

Issue #86, opened 2016-08-29
Bunching of Rows on Initial VScroll load


Tempus35:
On Initial load of Virtual Scrolling the rows all bunch up as one row and only move out after scrolling some.

![bunchedrows](https://cloud.githubusercontent.com/assets/1966467/18070645/1bc2a41e-6e04-11e6-8d11-053e095f9290.png)

amcdnl:
Can you make a repo case?

Tempus35:
Looking to make a plunker test or something, you think this has to do with the fact that each cell references a template in which the template loads a component, all of this thus would take a bit to load before showing in the table. IE the tables calculations for size might not detect right.

jimitndiaye:
I'm seeing the same problem with a minimal number of rows (about 5).

amcdnl:
Can you guys make a demo? Happy to look.

RicardoVaranda:
#121 seems like the same issue

dixso:
Hi,
You should initialize the data after the view init.

Example:
`
ngAfterViewInit() {
this.rows = [{foo: 'bar'}];
}
`

Regards!

amiller29au:
@dixso - Initializing in ngAfterViewInit did not work.

Duplicate issue here: [https://github.com/swimlane/angular2-data-table/issues/127](url)

dixso:
Hi @amiller29au 
If you use in the @component the property template, I think you should add _AfterContentChecked_ lifecycle.
I think the problem is the load time...

Example:
https://plnkr.co/edit/emZt6RUpsej2RLgfBuQd?p=preview

amcdnl:
@amiller29au fixed this :)!

omouse:
I'm having a similar issue to this right now, the body of the datatable is set to a height of 19px but the row height is set to 50.

the `ng-reflect-body-height` is set to "19", not sure why this is getting calculated incorrectly.

my code looks like this:

```
<ngx-datatable
   class="material"
   [rows]="rows"
   [columns]="columns"
   [loadingIndicator]="loading"
   [scrollbarV]="true"
   [rowHeight]="50"
   [headerHeight]="50"
   [footerHeight]="50"
   ></ngx-datatable>
```

and this is the problem:

![too-small-table](https://cloud.githubusercontent.com/assets/2110497/22411665/45a2a5fc-e674-11e6-8a93-83f7678d9bf8.png)

PR #156, opened 2016-09-28
Fix bunching of rows on Initial Load when using VScroll

amiller29au:
This solves the issues mentioned in #86 and #127 thanks to @marjan-georgiev for the initial investigation work to help figure this one out.


amcdnl:
Thanks!!

PR #157, opened 2016-09-28
Fix issue with StateService not being included in constructor

amiller29au:
State service is missing in constructor


amcdnl:
Fixed already :)

PR #167, opened 2016-10-01
Ensure minWidth and maxWidth values are specified saved as numbers

amiller29au:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Column minWidth and maxWidth specified in html are stored as strings, so when the calculation of the total widths are done it is string concatenation not summing values. 

**What is the new behavior?**

Convert minWidth and maxWidth to number when the values are set so that total width calculation works correctly.

- [ ] Yes
- [x] No

**Other information**:


istiti:
+1

PR #177, opened 2016-10-03
Fix undefined property name error in DataTableHeader

jtomaszewski:
Current master has a serious bug. https://www.dropbox.com/s/ket7376jr2ff8os/Screenshot%202016-10-03%2023.23.36.png?dl=0


amcdnl:
Fixed in 0.9.2 :)

PR #155, opened 2016-10-03
Column resize sometimes gives weird behaviour on mouse resize/click

jtomaszewski:
If you resize column with a mouse click & drag a couple of times, sometimes it jumps to a weird position or doesn't "unclick" at all. More details: https://drive.google.com/file/d/0B5-IASgGnqwdN214NmVrWFkzb3M/view (problems visible from ~0:20). 

After a couple of resizing clicks, the sort button stops working as well - you have to click on it multiple times, sometimes the click resizes the column instead of changing sort direction - etc.

Any ideas what exactly can be wrong? ;s


jtomaszewski:
Fixed in the commits above. Ready to merge.

jtomaszewski:
O, sorry, not yet, its not fixed. The problem appeared when `event.movementX` was equal to 0 and so then variable jumped to `(event.screenX - this.prevScreenX)`. I'll make push force better commit in a sec.

jtomaszewski:
Now it's fixed.

jtomaszewski:
Sorry for spamming. Previous way just wasn't good and was buggy. Now I simplified it and it works 100%.

amcdnl:
You're awesome! :)

PR #185, opened 2016-10-05
Add visible property to column to make it simpler to show/hide

amiller29au:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Difficult to change column visibility, need to add/remove from table options

**What is the new behavior?**

Set visible property on table column

- [ ] Yes
- [x] No

**Other information**:


amcdnl:
Hmmm ... I see the use case but on the fence about it.

amcdnl:
Since this can be accomplished via the current API, I'd rather not add more options to confuse people.

PR #189, opened 2016-10-08
(chore): Improve the build performance when doing development

java2kus:
- [ ] Bugfix
- [ ] Feature
- [X] Build related changes
- [ ] CI related changes

Currently when doing development using the Angular2 Data table code base, it takes longer to do incremental builds (HMR). 

**What is the new behavior?**
The initial as well as incremental build process time has improved significantly.   On my Macbook pro, the incremental build time has come down from 9-10 seconds to approx 2 seconds per change.   Initial boot time has also improved from 18 seconds to 8 seconds.    This has been achieved by making the following changes:

- Package all vendor dependencies into a DLL.
- Made a change to tsconfig to not emit declaration data and have set both forkChecker and usewebPackText to true thus significantly improving performance.

- [X] Yes
- [ ] No

In order to do development in HMR mode,  we need to do the builds using 

```
npm run build:dll
npm run start:hmr
```

After the dll is created in the dist folder, the webpack dev server can be run in HMR mode.


amcdnl:
@java2kus -

- I had turned ForkChecker off on purpose as it was broken at one point. It has since been fixed I just forgot to turn it back on.
- Whats the difference between DLL and CommonChunk? I was under the impression CommonChunk did what DLL is described to do.

java2kus:
@amcdnl 

- Thanks for clarification on ForkChecker part.   I observed that turning it and the usewebPackText on improves the performance a lot.
- Both CommonChunk plugin and DLL plugin are used to split code dependencies into chunks thus providing performance benefits.   But when it comes to hot reloads and incremental builds, webpack processes all the javascript files in the vendor and polyfill array irrespective of whether anything changed  for CommonChunks plugin.   For DllPlugin, webpack just processes the DLL references just once i.e. first time load and on each subsequent reload, this whole chunk of dependencies (e.g. Angular, socketjs (used by webpack dev server) etc.) are skipped because they are already included statically in index.html.

I use the above trick for my application's dev build where I have included loads of dependencies (angular, ng material 2, ng bootstrap etc.)  and this has reduced the incremental build times significantly to around 3 seconds from 35s.   Do try it out on your machine and let me know the results.

java2kus:
@amcdnl I botched up my git repository.  If you are still interested in merging please do let me know and I will create a pull request again.

PR #191, opened 2016-10-08
Add deep value getter to sortRows function

### LINKED ISSUE #181: Default sort doesn't work with "deep" properties
**I'm submitting a ...**  
```
[ ] bug report => search github for a similar issue or PR before submitting
[x] feature request
```

**Current behavior**
When you specify a "deep" property as the `prop` for a column: 

```typescript
new TableColumn({
    name: 'Cylinders',
    prop: 'engine.cylinders'
}),
```

or 

```html
            <datatable-column name="Cylinders" prop="engine.cylinders">
              <template let-value="value">
                {{value}}
              </template>
            </datatable-column>
```

Sort malfunctions, the entire data set is not sorted.

**Expected behavior**
All rows are sorted based on the value for the column. This is the behavior for rows that use a "shallow" property

http://plnkr.co/edit/9Eug63qi775HVCMe3BhO?p=preview

1. On either the "Expressive: Cars" tab or the "Templated: Cars" tab, try to sort by Formation
2. See that the sort does not work.
3. On either the "Expressive: Cars" tab or the "Templated: Cars" tab, try to sort by Color
4. See the the sort does work

I need to sort on deep properties, 

Since its trivial (from the usage standpoint) to display these properties using the `prop` input, it seems like sorting by them should be supported by the default sort 

Windows
VSCode
Node
Webpack Dev Server

* **Table version:** 0.7.x
0.9.2

* **Angular version:** 2.0.x
@angular/core@2.0.1

* **Browser:**
Only tried in Chrome

* **Language:** 

TypeScript@2.0.3

ashleahhill:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

#181 

**What is the new behavior?**
Default sort uses deepValueGetter to find values to compare.

- [ ] Yes
- [x] No

**Other information**:

Fixes #181

### DIFF
--- a/src/utils/sort.ts
+++ b/src/utils/sort.ts
@@ -1,6 +1,6 @@
 import { Sort } from '../models';
 import { SortType, SortDirection } from '../types';
-
+import { deepValueGetter } from './deep-getter';
 /**
  * Gets the next sort direction
  * @param  {SortType}      sortType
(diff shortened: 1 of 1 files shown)


amcdnl:
Thanks!!

PR #195, opened 2016-10-12
Removed package.json engine restrictions

ocombe:
**What kind of change does this PR introduce?**

The lib restricts the engine to node ~6.0.0 for no reason (this lib doesn't use any binary). This is a problem when using Yarn because it is more strict than npm and it will respect the engine restrictions which will fail the installation of all the npm modules with the following error message:
`error angular2-data-table@0.9.3: The engine "node" is incompatible with this module. Expected version "~6.0.0".`

**What is the new behavior?**
Removed the restriction

**Does this PR introduce a breaking change?**
- [ ] Yes
- [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,10 +32,6 @@
     "release:css": "node-sass -o release/ src/themes && node-sass -o release/ src/components",
     "release:minify": "uglifyjs release/index.js --source-map release/index.min.js.map --source-map-url release/index.js.map --compress --mangle --screw-ie8 --output release/index.min.js"
   },
-  "engines": {
-    "node": "~6.0.0"
-  },
-  "engineStrict": true,
   "repository": {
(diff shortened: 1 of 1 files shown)


amcdnl:
I added the engine restrictions to prevent people from downloading it with Node 4.x and it having build errors.

ocombe:
you only have build errors if you clone it from github and try to run it, not if you install it from npm right ?

amcdnl:
Correct.

ocombe:
Then the engine restriction in the package.json doesn't help, because it only works when you npm install it from npm :)

amcdnl:
Thanks! :)

niklas-dahl:
I'm using node v6.3.1 and I still get this error using yarn, is this intended?

PR #200, opened 2016-10-12
Fix #166 - Style sorted columns

### LINKED ISSUE #166: Highlight ordered columns
To better notice which column is ordered, a class could be added to the ordered column so it can be styled to the user's need.

I am currently investigating the best way of implementing this.

yann510:
- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

**What is the new behavior?**

Added class on sorted body cells

- [ ] Yes
- [X] No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -229,3 +229,4 @@ pip-log.txt
 
 #Mr Developer
 .mr.developer.cfg
+yarn.lock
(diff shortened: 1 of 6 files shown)


amcdnl:
I like this idea, curious why not just key off the current sort props instead of all new props?

amcdnl:
Added manually w/ my latest push

PR #218, opened 2016-10-24
(fix):  Fixed bug when ColumnMode.force is used to fill up viewport width.

java2kus:
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When maxWidth/minWidth are set for a few columns the calculation for columnWidths does not fill up the entire viewport.   Issue is in the below line which is executed only once:
 https://github.com/swimlane/angular2-data-table/blob/master/src/utils/math.ts#L116 

even though the delta width is not filled when maxwidth/minWidth is reached below:
https://github.com/swimlane/angular2-data-table/blob/master/src/utils/math.ts#L128-L131

**What is the new behavior?**
New behavior fills up the width greedily until the entire view port is filled.

- [ ] Yes
- [X] No

**Other information**:

…idth.

### DIFF
--- a/src/utils/math.ts
+++ b/src/utils/math.ts
@@ -98,40 +98,71 @@ function scaleColumns(colsByGroup: any, maxWidth: any, totalFlexGrow: any) {
  * @param {int} expectedWidth
  */
 export function forceFillColumnWidths(allColumns: any, expectedWidth: any, startIdx: any) {
-  let contentWidth = 0;
 
   let columnsToResize = startIdx > -1 ?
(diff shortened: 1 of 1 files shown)


java2kus:
@amcdnl  Forgot to attach the url to plunker for the bug above.   Find below the plunk showing the bug.   Run it in full preview mode to see that there is a lot of white space after the  last column ( i.e. state which should have been filled because of ColumnMode.force option) 

https://plnkr.co/edit/zpHimbcKl1qqMm0JKjtB

PR #275, opened 2016-11-10
fix for first click sorting. It used the previous value for the sorts…

gerhardboer:
- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When the user clicks on a column header to sort, the first click does not seem to do anything.
The second click sorts the rows.
see also https://github.com/swimlane/angular2-data-table/issues/254

**What is the new behavior?**
The rows are sorted on the first click of the column header.

- [ ] Yes
- [x ] No

**Other information**:

… when calling the sorting function. Changed it to use the sort that comes in from the event.


amcdnl:
Thanks!

PR #276, opened 2016-11-11
custom comparator does not update rows

gerhardboer:
- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When a custom comparator for a column is configured, the table is not sorted.

**What is the new behavior?**
A custom comparator should return a new array, which will be picked up by the change detection and update the view.

- [x ] Yes
- [ ] No

Custom comparators HAVE TO return a new array.
Not sure if it'll break anything, but it is an API change. Also added test.

**Other information**:

…detection can pick it up and update the view.


amcdnl:
Perfect.  Thanks for your contribution!

PR #285, opened 2016-11-15
fix(selection): init selection to empty array

ocombe:
- [x] Bugfix

If the input `[selected]` is not initialized, then clicking on a row when the option `[selectionType]` is set will trigger an error:
```
error_handler.js:54 TypeError: Cannot read property 'slice' of undefined
    at DataTableSelectionComponent.selectRow (index.js:1194)
    at DataTableSelectionComponent.onActivate (index.js:1213)
    at View_DataTableBodyComponent3.handleEvent_2 (component.ngfactory.js:454)
    at View_DataTableBodyComponent3.<anonymous> (view.js:418)
    at SafeSubscriber.schedulerFn [as _next] (async.js:101)
    at SafeSubscriber.__tryOrUnsub (Subscriber.js:223)
    at SafeSubscriber.next (Subscriber.js:172)
    at Subscriber._next (Subscriber.js:125)
    at Subscriber.next (Subscriber.js:89)
    at EventEmitter.Subject.next (Subject.js:55)
    at EventEmitter.emit (async.js:81)
```

**What is the new behavior?**
The selection should work even if `[selected]` has not be initialized

- [x] No

### DIFF
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -67,7 +67,7 @@ export class DataTableBodyComponent {
   @Input() detailRowHeight: any;
   @Input() emptyMessage: string;
   @Input() selectionType: SelectionType;
-  @Input() selected: any[];
+  @Input() selected: any[] = [];
   @Input() rowIdentity: any;
   @Input() rowDetailTemplate: any;
   @Input() selectCheck: any;


amcdnl:
@ocombe - do u think a null check would be better than a init?

ocombe:
That could do it too. It's not a regular use case anyway because there isn't a lot of sense in using selection type without selected.
I'm using it because they want to be able to track the lines in really large tables when they scroll horizontall. I don't really need the selection, just the fact that it highlights the lines

amcdnl:
Thinking on it more, I think empty is probably better as ppl might want to track 'focused' w/o any selections ( like your example ) and doing a falsey check would not facilitate that very well.

PR #286, opened 2016-11-16
Custom comparator should return new array

gerhardboer:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Custom sort functions do not update the reference to the rows, which means the table is not updated.

**What is the new behavior?**
The table is updated when using a custom sort function

- [ ] Yes
- [x] No

**Other information**:


amcdnl:
Awesome! Thanks for your help!

PR #298, opened 2016-11-20
Pagerclass property added

Drane:
Added a pagerClass property in cssClasses so that the pager css classcan be customized easily. 
(e.g. useful for enabling for bootstrap 3 pagingvia the 'pagination' class)

### DIFF
--- a/src/components/datatable.component.ts
+++ b/src/components/datatable.component.ts
@@ -68,6 +68,7 @@ import { scrollbarWidth, setColumnDefaults, translateTemplates } from '../utils'
         [offset]="offset"
         [footerHeight]="footerHeight"
         [totalMessage]="messages.totalMessage"
+        [pagerClass]="cssClasses.pagerClass"
         [pagerLeftArrowIcon]="cssClasses.pagerLeftArrow"
         [pagerRightArrowIcon]="cssClasses.pagerRightArrow"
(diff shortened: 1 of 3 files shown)


shattar:
Out of curiosity, is there a reason that this could not be handled by themes or a stylesheet?  It seems like an angular2 anti-pattern to configure styles via inputs like this.  It seems like this should be part of a custom theme.  However, I have not used bootstrap3, so maybe I could be schooled.

amcdnl:
Ya I agree with @shattar , what's your use case you couldn't custom style existing classes

Drane:
@shattar @amcdnl 
I agree this feels like being somewhat of an anti-pattern.
But the upside of CSS based "frameworks" like bootstrap is that they offer you nice theming out-of-the-box. Recreating those frameworks for every 3rd party component, by copying the corresponding styles to the components custom class-names, that also seems somewhat tedious.... 

Or am I overlooking something, and is it possible to translate class names somehow? That also seems interesting, and might be worth my time investigating it further then.

Thanks for the feedback and sorry if my pull request didn't make much sense.  Not very experienced in contributing pull requests (yet!). :)

amcdnl:
You want to use the bootstrap css styles? This isn't the same dom structure as bootstrap so it wouldn't really work. I also do not want to add those dependencies.

PR #314, opened 2016-11-23
Set default value for selected

ghetolay:
- [X ] Bugfix

Selection ins't working because `selected` property of DataTableComponent is not initialise, then that property is passed down to `DatatableBodyComponent` and `DatatableSelectionComponent` through `@Input()`. So even if both `DatatableBodyComponent` and `DatatableSelectionComponent` have an initial value for `selected` it gets overridden.

**What is the new behavior?**
It works !

**Does this PR introduce a breaking change?**
- [ ] Yes
- [X] No

Again, I've done that directly on github, tell if you want me to clone and build it.

### DIFF
--- a/src/components/datatable.component.ts
+++ b/src/components/datatable.component.ts
@@ -122,7 +122,7 @@ export class DatatableComponent implements OnInit, AfterViewInit {
   }
 
   // Selected rows
-  @Input() selected: any[];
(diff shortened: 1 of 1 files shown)


amcdnl:
@ocombe - didn't u fix this?

Thanks for your pr!

PR #328, opened 2016-11-30
New selection implementation

ghetolay:
**What kind of change does this PR introduce?**
- [x] Refactoring with small api changes (intended).

## Intro
This PR is not complete and if accepted should be the first of a serie of refactoring.

This project relies a lot on passing down and up the tree `@Input()` and `@Output()` which makes huge components with a lot of stuff not directly related to the component. I think we should refactor the code and use services, directives and injection to keep components clean.

Also I needed an API to select rows programmatically so I decided to refactor the selection mechanism.

## Refactor
- Remove the SelectionComponent and so the corresponding element on the dom.
- Remove all the selection related input/output from all components except for `DataTableBodyRowComponent` which is the only component using it.
- Create a selection directive that users must declare along with the datable (as if it was an input).
- Create an API on that directive to be used by other internal component as well as user.
  - `selection`: returns a copy of the current selected items.
  - `select(row, state): select or unselect the row depending on state. (if no state toggle row state).
  - `selectRowsBetween(firstRow, secondRow)`: to select multiple adjacent row (used for multiShift).
  - `toggleSelect(row, range)`: toggle a row or a range of row. Primarily used to react from events (click, enter...).
   - `selectAll(state)`: select or unselect all rows depending on state.
   - `isSelected(row)`: returns true if row is selected.

## Usage
Since the selection directive must be declared at the same level as `DatatableComponent` like that : 
```html
<datatable selection="multi"></datatable>
```

User can directly set inputs and listen to outputs like you would do for datatable. He can also inject `SelectionDirective` to interact directly with it.

```html
<datateble selection rowIdentity="" (selectionchange)=""></datatable>
```

```ts
//if datatable is a child of the component
@ViewChild(SelectionDirective) selection;

//if datatable is a parent of the component
constructor(@Optional() private selection: SelectionDirective){ }
```

Another benefit is if user doesn't specify the selection attribute, the directive won't be instantiate so no extra processing.

- [x] Yes
- [ ] No

I've made the following breaking changes because I think it's better, we could discuss about that : 

input: `selectionType` => `selection`
output: `select` => `selectionChange`

**Other information**:
This PR is not complete, this is just the first step. 
Here is what's missing : 

- I didn't migrate yet the focus behavior so it's currently not functional.
- I don't know how multiShift is supposed to work, I'll need some spec. Currently all tests for multiShift fail.
- Nesting datatable could produce bug since we can't use `@Host()` on the directive to contain the injection on `DataTableBodyRowComponent`. `DataTableBodyRowComponent` should be a direct child of `DatatableComponent`.

If you're willing to refactor the code to work like this new implementation I could open issues/PR to discuss about ideas of doing so.


amcdnl:
@ghetolay - First off, thanks so much for this! We <3 when we can get community feedback and code is even more awesome!

So a few notes:

> This project relies a lot on passing down and up the tree @Input() and @Output()

This is a necessary evil when using `OnPush`, when you include a service like that it can't use that change detection strategy since it doesn't know about changes outside of its internal scope. 

I actually had it using a service in a pre 1.0 release but refactored it to be all push which requires each component to have a Input immutable update to trigger the update. `OnPush` is similar to how React works, you will notice looking at those type components they all pass down their data. 

Also making it push down all the data via inputs makes it a totally isolate testable component at every micro level. No component knows about its parent.

> ...which makes huge components with a lot of stuff not directly related to the component.

TBH, I didn't think it was that bad. I split every piece into its own component/directive. Yes, there is some overlap with passing models around because of some limitations in Angular framework. Here is one particular bug pending for some of that cleanup: https://github.com/angular/angular/issues/4062

My last question is what does this accomplish that is not currently available? You mentioned you need a API to select rows, thats actually rather easy...take the following example:

```
import { Component } from '@angular/core';

@Component({
  selector: 'multi-selection-demo',
  template: `
        <datatable
          class="material"
          [rows]="rows"
          [columnMode]="'force'"
          [columns]="columns"
          [headerHeight]="50"
          [footerHeight]="50"
          [rowHeight]="'auto'"
          [limit]="5"
          [selected]="selected"
          [selectionType]="'multi'"
          (select)='onSelect($event)'>
        </datatable>
  `
})
export class MultiSelectionComponent {

  rows = [] = someDataFromAPI;
  selected = [];

  columns: any[] = [
    { prop: 'name'} , 
    { name: 'Company' }, 
    { name: 'Gender' }
  ];

  onSelect({ selected }) {
    this.selected.splice(0, this.selected.length);
    this.selected.push(...selected);
  }

  externalSelectUpdate(rowIndexs: any[] = [ 4, 10, 22 ]) {
     this.selected = this.rows.filter((row, idx) { return rowIndexs.indexOf(idx); });
  }
}
```

You can see in that demonstration how I have a external function that will set the selected row indexes based on a array of select row indexes. Thats the power of `OnPush`, no fancy APIs are needed just push your changes down.

Does that solve your questions?

ghetolay:
> This is a necessary evil when using OnPush, when you include a service like that it can't use that change detection strategy since it doesn't know about changes outside of its internal scope.

Well that why I used a directive and not a service, directive have input and react to input change. But you can still use service along with `markForCheck()`.

With this implementation I think I've reach the exact same functionality without all the inputs/output on components. And I kept the `onPush`. I have to agree I change a bit the how event are handled and I did not integrate focus yet.

> My last question is what does this accomplish that is not currently available?

Well I was able to refractor this : 

```ts
    if(row.length == 0) {
       let newSelected = this.parentTable.selected.filter( r => r!= this._row);
       if(newSelected.length != this.parentTable.selected.length)
         this.parentTable.selected = newSelected;
       }
      else if( !this.parentTable.selected.find( r => r == this._row) )
	this.parentTable.selected = [...this.parentTable.selected, this._row];
```
and this 
```ts
if( selection.selected.find( (r: any) =>r == this._row) ) {
    if( this.subtable.selected.length < this._row.studies.length ) {
	let newSelection = this._row.studies.slice();
	this.subtable.selected = newSelection;
	this.cd.markForCheck();
    }
}
else
 if(this.subtable.selected.length > 0) {
   this.subtable.selected = [];
   this.cd.markForCheck();
}
```

into 
```ts
this.parentSelection.select(this.row, selection.length != 0);
```
and
```ts
this.subtableSelection.selectAll( this.parentSelection.isSelected(this._row) );
```

And I think it's neat because that component shouldn't manage the data table selection itself. This job should be done by the data table.

This PR is meant to see if we agree on what should be the future implementation and if we don't I'm totally fine with it. I didn't talked about all the changes I would like to apply because I wanted to keep it close to the PR content. If you want we could discuss about it, dunno if this PR is the right place.

In short I really believe we can achieve the same level of performance (if not better) and keeping the same features using services, concise components, less DOM wrapper element, and without mutating the rows data passed by the user.

amcdnl:
What would the implementation code look like?

ghetolay:
Well I don't know exactly but 
 - Only keep Input/output directly used by the component.
 - Behavior features will mainly be top level directives (like selection, scroller) and inject it on component who needs it. 
- I would use a wrapper object for rows and `ImmutableDiffer` to detect changes on the rows. (Right know I think that if you select a row and then remove it from the rows array it will be selected forever).
- Probably using pipes for filtering/pagination/virtual scroll

I may totally be wrong in some points (I hope not in all), but I would definitely try things that way.

amcdnl:
Side note: there is a lot of changes going on right now in a branch will be doing a 2.0 release this wk. You can see it here - https://github.com/swimlane/angular2-data-table/tree/observables ... some of this might need to change there

amcdnl:
The more I think about this, the more I agree selection is murking in the way but I'm not sure how to abstract this given complexity. Have you had any other thoughts?

amcdnl:
Any updates @ghetolay

ghetolay:
What's the problem with the implementation design of this PR ?

ghetolay:
What's missing (and is also missing with the current implementation) is a hook when a row is removed from the model to also remove it from the selected row.

amcdnl:
Closing due to inactivity.

PR #331, opened 2016-11-30
rowContextmenu output

isaacplmann:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

No way to listen to the `contextmenu` event for a particular row.  You can wire up custom cell templates, but that doesn't capture right-clicks on the padding between cell contents.

**What is the new behavior?**

There is a new `rowContextmenu` synchronous EventEmitter that sends out the `contextmenu` event and the row data.  The EventEmitter needs to be synchronous so that the listener can run `event.preventDefault()` before the browser's default context menu has appeared.

This doesn't actually implement a context menu feature - it just allows other context menu libraries (like my own [angular2-contextmenu](https://github.com/isaacplmann/angular2-contextmenu)) to be used with it.

See the Basic: Context Menu demo for an example of usage.

- [ ] Yes
- [x] No

### DIFF
--- a/demo/app.component.ts
+++ b/demo/app.component.ts
@@ -22,6 +22,7 @@ import { Component } from '@angular/core';
               <li><a href="#" (click)="state='hidden'">Hidden On Load</a></li>
               <li><a href="#" (click)="state='live'">Live Data</a></li>
               <li><a href="#" (click)="state='rx'">RxJS</a></li>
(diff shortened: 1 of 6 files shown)


amcdnl:
Awesome! Thanks for your contribution!

PR #342, opened 2016-12-02
Use consistent naming, rename datatable to data-table

### LINKED ISSUE #341: Naming consistency Datatable vs DataTable
```
[*] bug report => search github for a similar issue or PR before submitting
[ ] feature request
```
I see that in some places Datatable is used and other places it is DataTable (two words). I think it would be nice to make it consistent. Also the directive and component selectors should follow this, ie (swui-datatable -> Datatable or swui-data-table -> DataTable). Some IDEs show Datatable is not a word so it would be nice to use DataTable.

**Current behavior**

**Expected behavior**

* **Table version:** 0.7.x

* **Angular version:** 2.0.x

mmrath:
- [x] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

Fixes #341


amcdnl:
@Brocco - thoughts? ^

Brocco:
I think it is an acceptable mindset to consider "datatable" to be a single word (at least in code). With that being said, I think the renaming in this PR is not necessary given that the root selector (`swui-datatable`) does not have a separating dash. I'd feel differently if the root selector were to be `swui-data-table`.

Just my $0.02

amcdnl:
My thoughts @Brocco .... thanks for ur suggestions @mmrath

PR #355, opened 2016-12-08
fix(virtual-scroll): fix virtual scroll height calculation

deeg:
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When using external server side paging & virtual scroll, the height calculation was using the current loaded count, instead of the total count.

This had the behavior where as you scrolled down the scroller got smaller, instead of always being the same size and being in the correct location.

**What is the new behavior?**

It now has the correct virtual scrolling calculation.

- [ ] Yes
- [X] No


deeg:
After playing with this a bit more this seems to *mostly* work, but have some issues.

Now as I scroll down the pages load like normal, and everything looks good. 

It seems to be when I scroll to fast, or click and drag the scroller to an arbitrary position. In older versions of the library, the data that comes back from the call would be rendered at the correct visible position. It now seems that the server is still making a call with the correct offset, but once the call comes back you can't see the data.

I will look into this a bit more, but any suggestions are welcomed. You might want to hold off on merging this PR.

src/utils/row-height-cache.ts
```diff
@@ -30,7 +30,7 @@ export class RowHeightCache {
    * @param rowHeight The row height.
    * @param detailRowHeight The detail row height.
    */
-  initCache(rows: any[], rowHeight: number, detailRowHeight: number): void {
+  initCache(rows: any[], rowHeight: number, detailRowHeight: number, totalCount: number): void {
```
deeg: Will add to docs

deeg:
Not sure why Travis failed. Tests are passing on my local machine. Might need to be re-run.

java2kus:
@deeg @amcdnl  - For server side pagination with virtual scroll, we need to make sure that the row heights array cache is not getting initialized for a large number of rows. e.g. if you have a million rows on server then an array size of million would be initialized on the browser.   I am not sure whether thats the best way to handle this problem.   The best way is to have an alternate logic where the offset gets calculated based on a formula.   What this would mean is we would loose the detail row expansion state across server pages (may be cache a few page states).  This would be a better compromise rather than a row cache.

deeg:
@java2kus, that makes sense, I didn't realize the implications.

Is losing the detail row expansion state across pages a big deal? I think if you have moved on past a few pages, coming back to that page with the detail state being closed isn't a huge deal.

Was there some sort of calculation like this before (version 0.9.3) where this used to be working? Something I might be able to go back and reference?

I don't mind making a PR for the new approach if this is the way we should go. Do you have an idea of where this logic should live?

java2kus:
@deeg I am not sure about the exact version but before the row detail feature was introduced,  a direct calculation was being done to calculate the virtual row offset on the screen.   

Please see the PR - [201](https://github.com/swimlane/angular2-data-table/pull/201/files) for the logic before the role detail was introduced.

deeg:
@java2kus thanks for all the info.

I got sidetracked on some other stuff at work, but I hope to get back to working on this feature soon.

amcdnl:
Lets close this for now and re-open when u can touch on it.

Issue #386, opened 2016-12-21
CPU usage after sorting


dae721:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here
```

**Current behavior**
I only started looking at this component a couple days ago, but from my outsider's perspective this appears to be a bug. As you sort and/or resize columns, the CPU usage gradually creeps up (and stays there). I'm using Chrome on Windows 10, but I also repro'ed this on an older Win 7 machine using both IE and Chrome (and on that machine, the CPU usage quickly reached 50%).

**Expected behavior**
CPU to go back down to 0 after the sort completes

**Reproduction of the problem**
Bring up the Fluid Row Heights sample:
http://swimlane.github.io/ngx-datatable

Bring up Chrome's Task Manager (Shift+Esc, or via the button in the upper right corner/more tools). You can also see this happening in Windows Task Manager, but Chrome's task manager shows CPU usage on a per tab basis. Notice the CPU for the datatable tab is at 0 or near 0.

Now click on column headers to sort the rows. After each click the average CPU usage increases slightly. After a dozen or so clicks it's at about 15%. Go to another tab and it falls to about 1%. It jumps back up after coming back to the datatable tab.

Seems like it's related to the number of rows. If I go to the Inline Editing sample with 100 rows, 5 showing at a time, CPU load goes down to about 5%. If I go to the 10k rows sample, it goes up to about 25%.

Same thing happens if I resize the columns instead of sorting them, though doesn't seem to be as severe.

**What is the motivation / use case for changing the behavior?**
Fan eventually kicks in, which is when I first noticed the problem.

**Please tell us about your environment:**
Windows 10, Windows 7

* **Table version:** 4.0.0
I assume the online demo is using latest, but I was also able to repro on a sample app on localhost using 4.0.0.

* **Angular version:** 2.3.1

* **Browser:**  Chrome 55 | IE 11

* **Language:** TypeScript 2.1

dae721:
[...]

PR #409, opened 2017-01-03
Bug 386

gerhardboer:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

There are some performance issues which took an impact to the cpu.
According to https://github.com/swimlane/ngx-datatable/issues/386#issuecomment-268909728 this should fix the issue.

**What is the new behavior?**
I could not reproduce the cpu issue, but i did validate the code with a test. The 'longpress' class is now added, and it makes the text color of the column green.

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/src/directives/long-press.directive.spec.ts
+++ b/src/directives/long-press.directive.spec.ts
@@ -35,7 +35,7 @@ describe('LongPressDirective', () => {
     });
   }));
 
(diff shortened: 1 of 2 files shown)


amcdnl:
Looks good! Thanks as always!

PR #473, opened 2017-01-28
Fix bug 459

gerhardboer:
- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Cannot find file 'datatable.component.scss' and the AoT compilation fails.

**What is the new behavior?**
AoT compilation does not fail anymore

- [ ] Yes
- [ x] No, but

**Other information**:
It isn't pretty. I'm not sure why the relative path loading does not work. It might be an issue with a webpack loader or something.

I've tested before and after the change with an angular-cli app and an angular-class-starter app. Both had the same issue, and this kind of path loading fixes is.


amcdnl:
You shouldn't have to reference the scss like that :S .... its assuming you have scss running in your stack.

I think a simple replace ( ugh ) might fix this in the build step.

gerhardboer:
Yeah, i'm not happy about it either.

I actually expected the release version to NOT have any scss dependencies, or at least have the path of the styleUrls be wrapped in a webpack require, where the url it requires matches the one that it built.

It's the reason i tried this "solution", to make the path match more with the one it registers for the scss.

I don't have the webpack-knowledge to figure it out unfortunately.

amcdnl:
This is actually ngc not webpack...

gerhardboer:
Ah, well.. explains why nothing worked when i was fiddling with webpack.

Do you have a solution for this, or at least know where to look?

amcdnl:
I temp fixed the issue, need to figure out better way tho.

PR #492, opened 2017-02-07
longpress class not cleaned up

gerhardboer:
also reduced mouseup events from DraggableDirective and general fixes.

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

If you hold a column header to drag, the longpress class gets active. If you then move the cursor out of the element while holding the element and let go, the mousup event of the element does not get fired.

This holds the longpress class on the header element.

**What is the new behavior?**
The mousup handler gets called when there is a longpress active and the mousebutton is let go.

- [ ] Yes
- [x] No

**Other information**:
I couldn't find an issue for this, and since opening an issue would be more work than the fix was I did not create one. But if of course I could open one if you'd like.

### DIFF
--- a/src/directives/long-press.directive.ts
+++ b/src/directives/long-press.directive.ts
@@ -26,7 +26,7 @@ export class LongPressDirective {
   get press(): boolean { return this.pressing; }
 
   @HostBinding('class.longpress')
-  get isLongPress(): boolean { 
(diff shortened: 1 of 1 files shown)


gerhardboer:
might be the cause of a bug, will investigate more.

PR #494, opened 2017-02-08
Numeric prop support (support array rows)

arlowhite:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Column prop cannot be a number.

**What is the new behavior?**

numeric column props now work.

- [ ] Yes
- [x] No

**Other information**:

EDIT: Please reply on the code-review comments regarding these issues.

I added getter functions that are more efficient than the original deepValueGetter. deepValueGetter is only used if the prop string contains '.'

In my second commit, I cache the value getter function in **DataTableBodyCellComponent**, but this may be over-optimizing. However, this should improve performance if the value getter is called much more often than the column setter.

**Items to Review**

I would like to review the API for the getter functions. I'm not a fan of deepValueGetter's API as far as undefined values. Sometimes it returns '', other times it returns the object. To be consistent, I think all getters should return '' (the empty string) in undefined situations. This also lets you get rid of the undefined check in DataTableBodyCellComponent.


arlowhite:
This is another issue, but I think the sorting code could be much more efficient if it avoided valueGetter lookups in the compare function. Other aspects of the compare function body should be simplified too.

arlowhite:
As an alternative to baf9e82 (valueGetter caching in **DataTableBodyCellComponent**), the valueGetter function could just be stored in the column object. So the valueGetter would be looked up when columns are set. `this.column.valueGetter`

Would you prefer this approach?

arlowhite:
There's a bug right now. zero-index `prop=0` isn't working because code elsewhere doesn't render the column because of its falsey prop value. Working on it...

src/components/header/header-cell.component.ts
```diff
@@ -84,6 +84,10 @@ export class DataTableHeaderCellComponent {
 
   @HostBinding('attr.title')
   get name(): string {
+    // TODO review this.column.name logic. just return this.column.name?
+    // column name is guaranteed by setColumnDefaults()
+    // so this.column.prop is probably never returned
+    // However, if name is not truthy, this.column.prop could be a number and return
     return this.column.name || this.column.prop;
```
arlowhite: Any reason not to change this to:
`return this.column.name`

Otherwise, to be safe (and conform to the API) it should be:
`return this.column.name || String(this.column.prop)`

src/types/table-column.type.ts
```diff
@@ -1,6 +1,13 @@
 import { PipeTransform } from '@angular/core';
 
 /**
+ * Column property that indicates how to retrieve this column's
+ * value from a row.
+ * 'a.deep.value', 'normalprop', 0 (numeric)
+ */
+export type TableColumnProp = string|number;
```
arlowhite: Is **TableColumnProp** a good name for this type? Just **ColumnProp** instead?

src/utils/column-helper.ts
```diff
@@ -8,22 +9,23 @@ import { camelCase, deCamelCase, id } from '../utils';
...
  * @returns
  */
-export function setColumnDefaults(columns: any[]) {
+export function setColumnDefaults(columns: TableColumn[]) {
   if(!columns) return;
 
   for(let column of columns) {
     if(!column.$$id) {
       column.$$id = id();
     }
```
arlowhite: How about just determining the valueGetter function in this code instead of the caching I did in **DataTableBodyCellComponent**?

Is it safe to assume setColumnDefaults is always run on columns?

src/utils/column-prop-getters.ts
```diff
@@ -0,0 +1,87 @@
...
+  if (value == null) return '';
+  return value;
+}
+
+/**
+ * Returns a deep object given a string. zoo['animal.type']
+ * @param {object} obj
+ * @param {string} path
+ */
+export function deepValueGetter(obj: Object, path: string) {
```
arlowhite: IMO this function has too many possible return types for the missing cases.
**obj: Object** if `!obj || !path`
**empty string** if searching for the deep value fails.

Is there any reason for the `!obj || !path` condition not to `return ''`?

src/utils/column-helper.ts
```diff
@@ -16,14 +16,15 @@ export function setColumnDefaults(columns: any[]) {
...
 
+    // prop can be numeric; zero is valid not a missing prop
     // translate name => prop
-    if(!column.prop && column.name) {
+    if(column.prop == null && column.name) {
```
amcdnl: I'd prefer `=== undefined` vs `null`

src/utils/column-helper.ts
amcdnl: I'd say its safe to assume.

src/types/table-column.type.ts
amcdnl: Agreed.

src/components/header/header-cell.component.ts
amcdnl: Agreed.

src/utils/column-helper.ts
arlowhite: If we're certain that `column.prop` is `number|undefined`, then that's fine. I just did it because the original `!column.prop` would have caught `null|undefined|0`, but with numeric 0 being valid, I wanted to catch `null|undefined`

Maybe `setColumnDefaults` should throw an error on a `null` `column.prop` to enforce the assumption?

src/utils/column-helper.ts
arlowhite: Ok, I think it's better design to put the valueGetter on TableColumn instead since properties like `$$id` are already calculated in `setColumnDefaults()`

deeg:
I'm confused at what functionality this actually brings. What advantage do you get by allowing a prop to be a number?

arlowhite:
@deeg Support for array rows.

```javascript
this.rows = [
['Foo', 123],
['Bar', 456]
];

this.columns = [
 { name: 'First column', prop: 0},
 { name: 'Second column', prop: 1}
];
```

deeg:
@arlowhite, thanks for the info. I now understand the functionality it brings, but not sure it is something that has to be supported by this library.

What is the real life use case for this? You could very easily make that data work with this library currently my formatting the data before passing it into the table.

I'm not against the change, I just think it is adding a lot of code & logic for something that isn't a huge problem right now. I of course am just giving my opinion and will yield to the maintainers of the library and the community input.

arlowhite:
I suppose the argument is just to reduce the amount of work users need to do to use the table.

Yes, conversion of rows isn't that difficult, but it does add complexity. There are a few other issues.

- Users also need to convert back to their Array format (to be compatible with their code) when dealing with output such as from `(contextmenu)`.
- Users can no longer compare rows between their code and ngx-datatable by reference when their code holds a reference to the Array row form and ngx-datatable rows are different JSON objects.

arlowhite:
Just FYI, I'll working on updating this branch and resolving conflicts next week. (March 6-10)

arlowhite:
I just did a rebase onto master.

arlowhite:
I'm finished working on this PR. I think it's good to go.

arlowhite:
I just rebased onto master again.
I've just updated my project to Angular 4 and was able test ngx-datatable in my project.
Also, npm start and npm test looks good.

src/utils/column-helper.ts
arlowhite: @amcdnl What should happen if `column.prop` is `null`?

amcdnl:
Merged!

Issue #503, opened 2017-02-09
Improve performance with ChangeDetectionStrategy


arlowhite:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[x] feature request
[ ] support request => Please do not submit support request here
```

**Current behavior**

A complex table (many rows/columns) will cause the browser to hiccup on Angular change detection, causing jerky animations and other performance issues.

**Expected behavior**

ngx-datatable should be optimized to do minimal Angular change detection.

**Reproduction of the problem**

I noticed this when I added a `<md-input>` (angular/material2), which has a floating label animation. With a ngx-datatable as a sister component, the animation is jerky.
Also, do a Chrome Dev Tools JS CPU profile with and without ngx-datatable on a page. You will notice much more time spent in AppView.detectChanges. 

Expand:
**AppView.detectChanges > ViewContainer.detectChangesInNestedViews** and you will see many ngx-datatable detectChangesInternal calls.

**What is the motivation / use case for changing the behavior?**

ngx-datatable should have good performance in Angular.

* **Table version:** 6.0.2

* **Angular version:** 2.4.6

* **Browser:** Chrome

* **Language:** TypeScript 2.1.5

arlowhite:
I added `changeDetection: ChangeDetectionStrategy.OnPush` to **DatatableComponent**, but it didn't improve performance much. So then I added **OnPush** to **DataTableBodyComponent**, **DataTableBodyRowComponent**, **DataTableBodyCellComponent** and performance improved significantly.

I was under the impression that only the root component would need OnPush, but this doesn't seem to be the case. I don't have much experience on ChangeDetection optimization; I've just read a bit on it.

One problem with this is that users may need to change how they input data into ngx-datatable. To trigger an update to rows, they would need to change the reference (make a copy of rows array). Alternatively, rows could also be an Observable (`any[] | Observable<any[]>`); from what I've read Angular change-detection is Observable aware. Could also add a `refresh()` method or something, which triggers change detection, but this seems ugly to me in the Angular world.

Hopefully someone who has worked on ChangeDetection optimization knows the best way to design this.

arlowhite:
I set most Components to **OnPush** and it seems to work fine:
https://github.com/arlowhite/ngx-datatable/tree/change-detection-only

However, this probably should go through more testing.
And as I mentioned, changing rows and columns Inputs may behave differently so documentation may need to be updated.

[...]

PR #507, opened 2017-02-11
Removed document:mouseup because it caused change detection too often.

gerhardboer:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

By hooking into NgZone I found out that the global mouseup listeners in the header directives caused full changedetection to run. The mousup handlers were safe by checking if the body should be executed, but the cd cycle without the OnPush caused some performance degredation.
This is noticable in animations.

Should also have a positive impact on #503 

**What is the new behavior?**
MouseUp listeners are now only added when the user clicked the header, and destroyed when the mouseup event is thrown.
This means the mouseup events on the document do not have a handler all the time, so zonejs does not kick of the changedetection when a mouseup event is thrown anywhere on the document.

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/src/directives/draggable.directive.ts
+++ b/src/directives/draggable.directive.ts
@@ -37,8 +37,9 @@ export class DraggableDirective implements OnDestroy {
     }
   }
 
-  @HostListener('document:mouseup', ['$event'])
   onMouseup(event: MouseEvent): void {
+    if (!this.isDragging) return;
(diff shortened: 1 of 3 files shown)


amcdnl:
AWESOME! Thanks !

Issue #500, opened 2017-02-09
Show selection count at footer


wndproc1:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[x] feature request
[ ] support request => Please do not submit support request here
```

**Current behavior**
If there are many rows, it is hard to identify how many rows are selected currently.

**Expected behavior**
It would be great if the currently selected row count is displayed at the footer, next by the total count.
Example:
23 selected / 340 total

**Reproduction of the problem**
N/A

**What is the motivation / use case for changing the behavior?**
See "Current behavior" and "Expected behavior" above.

**Please tell us about your environment:**
Windows7, VSCode, angular-cli, Tomcat 8.5, ...

* **Table version:** 6.0.2

* **Angular version:** 2.4.6

* **Browser:** all

* **Language:** all

deeg:
I believe this landed and can be closed. @amcdnl

PR #508, opened 2017-02-11
Feature 500

gerhardboer:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

See #500 

**What is the new behavior?**
The text 'X selected /' is added when a selectionType has been set. Supports translations.

**possible improvements**
I'm not setting the selectedMessage if there is no selectionType with the same logic to check whether selection is turned on.
This might be better if the developers could set this themselfs with another input property, but since this is an api addition, i thought i'd check first.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/components/datatable.component.scss
+++ b/src/components/datatable.component.scss
@@ -228,14 +228,23 @@
       width: 100%;
     }
 
+    .selected-count {
+      .page-count {
+        flex: 1 1 40%;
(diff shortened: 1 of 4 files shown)


wndproc1:
Thank you very much!

PR #510, opened 2017-02-13
fix: fix row width and styles

deeg:
This fixes an issue where changing the width of the table component after it is rendered the background of the table rows do not look cut off.

[You can see the bug in action here.](http://plnkr.co/edit/26L6dAYk1d7xHvvGAPTK?p=preview)

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

You can see the issue here that if you init a table in a component with a certain width, and then change the width to a larger width, the striped background does not extend across the entire table.
Even changing the window size which re-calculates the table does not help, which means calling re-calculate would not solve the issue.

**What is the new behavior?**

The row width is always correct even if it is changed after the table is rendered so the background color extends the entire row.

- [ ] Yes
- [X] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "@angular/platform-browser": "~2.4.5",
     "@angular/platform-browser-dynamic": "~2.4.5",
     "@angular/platform-server": "~2.4.5",
-    "@types/jasmine": "^2.5.41",
+    "@types/jasmine": "2.5.41",
     "@types/node": "^7.0.4",
     "angular2-template-loader": "^0.6.0",
     "autoprefixer": "^6.5.4",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -72,7 +72,7 @@
     "@angular/platform-browser": "~2.4.5",
     "@angular/platform-browser-dynamic": "~2.4.5",
     "@angular/platform-server": "~2.4.5",
-    "@types/jasmine": "^2.5.41",
+    "@types/jasmine": "2.5.41",
```
deeg: Hardcoded this version for now due to `npm run package` not completing due to [this bug](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/14579)

amcdnl:
Thanks as always!

Issue #496, opened 2017-02-08
Columns cannot be resized if their label overlaps resize-handle


arlowhite:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here
```

**Current behavior**

Column resize-handle cannot be grabbed when column text overlaps it.

**Expected behavior**

Should always be able to drag resize handle.

**Reproduction of the problem**

Go to the demo: http://swimlane.github.io/ngx-datatable/#
Drag a column smaller so that the column label does not fit.
Now you cannot drag the column again!

**What is the motivation / use case for changing the behavior?**

Should always be able to resize columns.

**Please tell us about your environment:**

* **Table version:** 6.0.2

* **Angular version:** 2.4.6

* **Browser:** Chrome, Firefox

* **Language:** NA

**FIX**

Just set **z-index** on `.resize-handle`
Not sure what value you should use though.

smoke:
I have noticed this too, way better fix would be to just put the .resize-handle element right after the content of the cell.
Having a z-index would cause more issues than it solves.
In a schematic representation (can't look the exact code right now, sorry) it is rendered like so:
```
<th><span class="resize-handler"></span><span class="header-content">Name</span></th>
```
->
```
<th><span class="header-content">Name</span><span class="resize-handler"></span></th>
```

amcdnl:
Resizable is added at the header component here - https://github.com/swimlane/ngx-datatable/blob/master/src/components/header/header.component.ts#L22 ... which renders here https://www.screencast.com/t/BlVc1LL4Rbg

smoke:
You may have noticed or not that I filled a pull request that fixes this https://github.com/swimlane/ngx-datatable/pull/521

PR #521, opened 2017-02-15
Fix resize handle overlaped by label

smoke:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

https://github.com/swimlane/ngx-datatable/issues/496

**What is the new behavior?**
The `.resize-handle` is positioned right after the content of the header, so that browsers properly lays it out over it - allowing the resizing to be usable in such cases.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/directives/resizeable.directive.ts
+++ b/src/directives/resizeable.directive.ts
@@ -1,5 +1,5 @@
 import {
-  Directive, ElementRef, HostListener, Input, Output, EventEmitter, OnDestroy
+  Directive, ElementRef, HostListener, Input, Output, EventEmitter, OnDestroy, AfterViewInit
 } from '@angular/core';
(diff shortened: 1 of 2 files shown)


src/directives/resizeable.directive.ts
```diff
@@ -1,5 +1,5 @@
 import {
-  Directive, ElementRef, HostListener, Input, Output, EventEmitter, OnDestroy
+  Directive, ElementRef, HostListener, Input, Output, EventEmitter, OnDestroy, ngAfterViewInit
```
amcdnl: Its `AfterViewInit`, not `ngAfterViewInit`

smoke: @amcdnl Thanks for noticing that! This is now fixed.

PR #538, opened 2017-02-17
Fix for orderable unsubscribing 

gerhardboer:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

reproducable on https://swimlane.github.io/ngx-datatable/# 

When toggling the Name column to off and then clicking the 'Company' column, a stacktrace pops up in the console.
`
Error: object unsubscribed
    at t.h (polyfills.js:1651)
    at new t (polyfills.js:1616)
    at t.next (polyfills.js:1609)
...
`

**Problem**
The KeyValueDiffer is used with an Array, which means it will use the index of the array to check for diffs.
When the Name column is removed the column-array us spliced, so no gaps occure.

This will cause the differ to think that the actual removed column was on the last index, since it went from [0, 1, 2] to [0,1]. The Company column was on this index, so it unsubscribed from the dragStart and dragEnd event emitters of the Company column, instead of the Name column.

**Fix**
By transforming the QueryElement Array into a map with the column.$$id as key, the KeyValueDiffer uses the column.$$id to diff the objects.

**What is the new behavior?**
The Company column does not lose its subscription to the draggable directive.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/directives/orderable.directive.spec.ts
+++ b/src/directives/orderable.directive.spec.ts
@@ -1,8 +1,10 @@
-import { async, TestBed, ComponentFixture } from '@angular/core/testing';
-import { Component, DebugElement } from '@angular/core';
-import { By } from '@angular/platform-browser';
+import {async, ComponentFixture, TestBed} from "@angular/core/testing";
+import {Component, ElementRef} from "@angular/core";
+import {By} from "@angular/platform-browser";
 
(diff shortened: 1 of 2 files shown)


amcdnl:
Awesome idea 👍

PR #540, opened 2017-02-20
location of changelog.md file

istiti:
- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ X ] Other... Please describe:

not fast tofind changelog.md need to go inside docs... 

**What is the new behavior?**
empalcement of changelog.md file

- [ ] Yes
- [ x] No


deeg:
I tend to agree that it should be a root level so it is easy for people to find. I know this may cause issues with it displaying on the docs page, not sure where it is pulling it from.

Maybe at least link to it in the readme, if you don't want to move the location?

amcdnl:
Ya, The reason its there is for displaying in the docs. Maybe a link in the readme might be better?

PR #541, opened 2017-02-20
Change output rowContextmenu to tableContextmenu.

frehn:
Instead of `{ event: MouseEvent, row: any }`, the new output emits an event of type `{ event: MouseEvent, type: ContextmenuType, content: any }`. If the event is triggered by a click on the body of the table, i.e. a row, then `type == 'body'` and `content` contains the clicked row. If the event is triggered  by a click on the header, then `type == 'header'` and `content` contains the column object of the clicked column.

It is useful if a contextmenu should be shown for the header of the table. For example, in a project of mine, I'd like to show a input box in a context menu that allows the user to filter the column whose header was right-clicked.

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

No events are fired if the header is right clicked.

**What is the new behavior?**
An event is fired via the tableContextmenu output when the table is clicked. For a more precise description, see above.

- [x] Yes
- [ ] No

```
onContextMenu(contextMenuEvent) {
  ... = contextMenuEvent.row;
}
```
becomes
```
onContextMenu(contextMenuEvent) {
  if (contextMenuEvent.type == 'body') {
    ... = contextMenuEvent.content;
  }
}
```

**Other information**:

### DIFF
--- a/release/components/datatable.component.d.ts
+++ b/release/components/datatable.component.d.ts
@@ -299,6 +299,15 @@ export declare class DatatableComponent implements OnInit, AfterViewInit, DoChec
         row: any;
     }>;
     /**
+     * The context menu was invoked on the header.
+     *
+     * @memberOf DatatableComponent
+     */
+    headerContextmenu: EventEmitter<{
+        event: MouseEvent;
+        header: string;
+    }>;
+    /**
      * CSS class applied if the header height if fixed height.
      *
      * @readonly
--- a/release/components/datatable.component.d.ts
+++ b/release/components/datatable.component.d.ts
@@ -299,15 +299,6 @@ export declare class DatatableComponent implements OnInit, AfterViewInit, DoChec
         row: any;
     }>;
     /**
-     * The context menu was invoked on the header.
-     *
-     * @memberOf DatatableComponent
-     */
-    headerContextmenu: EventEmitter<{
-        event: MouseEvent;
-        header: string;
-    }>;
-    /**
      * CSS class applied if the header height if fixed height.
      *
      * @readonly
--- a/src/components/datatable.component.ts
+++ b/src/components/datatable.component.ts
@@ -39,7 +39,8 @@ import { DatatableFooterDirective } from './footer';
         (sort)="onColumnSort($event)"
         (resize)="onColumnResize($event)"
         (reorder)="onColumnReorder($event)"
-        (select)="onHeaderSelect($event)">
+        (select)="onHeaderSelect($event)"
+        (headerContextmenu)="headerContextmenu.emit($event)">
       </datatable-header>
       <datatable-body
         [rows]="rows"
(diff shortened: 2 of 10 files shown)


release/components/datatable.component.d.ts
```diff
@@ -285,6 +285,15 @@ export declare class DatatableComponent implements OnInit, AfterViewInit, DoChec
         row: any;
     }>;
     /**
+     * The context menu was invoked on the header.
+     *
+     * @memberOf DatatableComponent
+     */
+    headerContextmenu: EventEmitter<{
```
amcdnl: Please don't include compiled files in the PR.

src/components/datatable.component.ts
```diff
@@ -452,6 +453,13 @@ export class DatatableComponent implements OnInit, AfterViewInit, DoCheck {
   @Output() rowContextmenu = new EventEmitter<{ event: MouseEvent, row: any }>(false);
 
   /**
+   * The context menu was invoked on a header cell.
+   *
+   * @memberOf DatatableComponent
+   */
+  @Output() headerContextmenu = new EventEmitter<{ event: MouseEvent, name: string, prop: string }>(false);
```
amcdnl: What do you think about consolidating row and header into a single event with a type?

amcdnl: Rather than pass the name and prop, lets just pass the whole column object. What do you think?

amcdnl: We should also make sure to update the header / row events to have similar arguments to the event if they don't already.

release/components/datatable.component.d.ts
frehn: Sorry, this slipped into the commit; I'll remove it.

src/components/datatable.component.ts
frehn: That's very reasonable.

My thoughts on having a single event:
- Nicer API due to less outputs.
- Not backwards compatible: `rowContextmenu` should be renamed to `contextmenu`, the event interface will be changed.

If backwards compatibility is not a strong requirement, I'd prefer to implement your suggestion.

frehn: That's very good - I don't know why I did not do it this way in the first place.

frehn: Sorry, I'm not sure I understand - could you clarify what you mean by this? I'm not sure what "the event" is referring to.

amcdnl: Non-backwards compatible is fine. Does `contextmenu` actually work? I've had issues in the past with event collisions.

amcdnl: Sorry I mean't the `output` arguments.

frehn:
I've updated the PR to address the requested changes.

frehn:
@amcdnl: I'd be really interested to get this merged. Is there anything I can do? What are your plans for this PR?

amcdnl:
Can you resolve the conflicts and i will merge :)

frehn:
Done - thank you :-)

amcdnl:
Merged. Thanks for your contribution!

frehn:
Thank you for the merge!

Issue #515, opened 2017-02-15
Bug with filtering data


ritikrishu:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here
```

**Current behavior**

Filtering data shows total count >0 but no rows are displayed.

**Expected behavior**

See filtered data.

**Reproduction of the problem**

In the demos, on basic => filtering. Go to page offset >3 type *el* in search. Footer shows 15 total entries but no entries are shown.

**What is the motivation / use case for changing the behavior?**

User can search to filter data from any page offset and table should display filtered data properly.

**Please tell us about your environment:**

VS code
* **Table version:** 0.7.x

6.1.2
* **Angular version:** 2.0.x

2.4.7
* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

all
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
all

ritikrishu:
PR: https://github.com/swimlane/ngx-datatable/pull/516 should resolve this issue

deeg:
#516 causes major issues with server side pagination and needs to be reverted, so this bug will need to be fixed another way.

amcdnl:
Per @deeg I have reverted that fix.

ritikrishu:
@deeg @amcdnl see if https://github.com/swimlane/ngx-datatable/pull/549 is a correct approach

deeg:
I am bringing the discussion back here so we can talk about the actual problem at hand.

From what I can tell we both have the same problem. 

**The steps to reproduce are as follows:**

* Navigate to page 4
* Filter your data for the table so that the remaining results have less than 4 pages (2 pages).
    * The mechanism of filter does not matter, could be a client side text filter or a server side facet which re-fetches data
* Now the offset of the table does not change, so you are 'looking at' page 4 when it doesn't really exist. 
* If you click on page one or two which has data, the filtered data would show correctly.

**Correct Behavior**

I believe when the user is filtering/faceting data they always want to go back to the first page. If I'm searing for everything with a certain term, I often want to start at the beginning of the list, not arbitrarily on a middle page.

**The Problem**

...is the library doesn't know if you are filtering data, or just loading a new page from server side pagination.

**Possible Solutions**
[...]

PR #549, opened 2017-02-22
Fix: Bug with filtering data & server side pagination

ritikrishu:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

issue https://github.com/swimlane/ngx-datatable/issues/515
associated issue from PR https://github.com/swimlane/ngx-datatable/pull/543

**What is the new behavior?**
If there is enough data to display, no change in `offset` is done. (cases: server side pagination or filtering)
If there are not enough rows to display, change `offset` to maximum value that can show table rows. (cases: filtering)

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/datatable.component.ts
+++ b/src/components/datatable.component.ts
@@ -790,8 +790,8 @@ export class DatatableComponent implements OnInit, AfterViewInit, DoCheck {
   recalculatePages(): void {
     this.pageSize = this.calcPageSize();
     this.rowCount = this.calcRowCount();
-    while(this.rowCount / this.pageSize < this.offset) { 
-      this.offset--;
+    if(this.rowCount / this.pageSize < this.offset) {
+      this.offset = Math.floor(this.rowCount / this.pageSize);
     }
   }
 
--- a/src/components/datatable.component.ts
+++ b/src/components/datatable.component.ts
@@ -790,6 +790,9 @@ export class DatatableComponent implements OnInit, AfterViewInit, DoCheck {
   recalculatePages(): void {
     this.pageSize = this.calcPageSize();
     this.rowCount = this.calcRowCount();
+    while(this.rowCount / this.pageSize < this.offset) { 
+      this.offset--;
+    }
   }
 
   /**
--- a/src/components/datatable.component.spec.ts
+++ b/src/components/datatable.component.spec.ts
@@ -13,6 +13,77 @@ describe('Datatable component', () => {
     TestBed.compileComponents();
   }));
 
+  describe('When data for rows change', () => {
+    it('should not alter offset if there is enough data to display in table', () => {
+      let fixture = TestBed.createComponent(DatatableComponent);
+      let initialRows = [
+        { id: 1 },
+        { id: 2 },
+        { id: 3 },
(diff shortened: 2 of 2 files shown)


src/components/datatable.component.ts
```diff
@@ -790,6 +790,9 @@ export class DatatableComponent implements OnInit, AfterViewInit, DoCheck {
   recalculatePages(): void {
     this.pageSize = this.calcPageSize();
     this.rowCount = this.calcRowCount();
+    while(this.rowCount / this.pageSize < this.offset) {
```
amcdnl: Can you explain what this does?

ritikrishu: suppose `rowCount` is `15`, `pageSize` is `10` that means table can have at max 2 pages(0 and 1 `offset` value) 10 rows in first and 5 rows in second). This was checking `while` current `offset` is greater than 1, then decrement it.
BTW, I have replaced the while loop with math function in https://github.com/swimlane/ngx-datatable/pull/549/commits/00eb63af1848b2635babd64a3440c5b58db2dcbd which directly checks if current `offset` is greater than number of required pages in table, then set `offset` to maximum possible page:
```
if(this.rowCount / this.pageSize < this.offset) {
       this.offset = Math.floor(this.rowCount / this.pageSize);
}
```
Here in example we took, suppose there are 15 available rows and limit is 10 then offset will be set to Math.min(15/10) => Math.min(1.5) => 1

ritikrishu: this is mainly to insure that server side pagination works fine as well filtering.

deeg:
I am now facing a simmilar issue when I want to use a facet to replace data which is in the table.

I'm finding that I want to go back to the first page after replacing my data. I have pulled down this PR but it doesn't quite work. I am going to see if I can come up with something that works for both of us.

I would maybe hold off on merging this until I have tomorrow morning to look at it if that is alright.

ritikrishu:
Yup you are right, that should still be a issue as this implementation assumes that you have your table with one set of data that can grow, or filtered but won't change in between.
Solution I can think of is instead of changing actual input (rows) from outside on some filter we can have inbuilt generic filters like of sorting. @deeg @amcdnl

ritikrishu:
> I want to use a facet to replace data which is in the table

BTW if you are meaning multiple filters here, then I think this should work fine as this has nothing to do with implementation of filters, it just plays with the new data set passed.
If you still find something non-functional, please share I'll also try to find a rigid solution to this. @deeg

deeg:
@ritikrishu, please see the discussion back on the issue page.

amcdnl:
@deeg - Any progress on this?

deeg:
@amcdnl, take a look at [my comment here](https://github.com/swimlane/ngx-datatable/issues/515#issuecomment-282058754)

I went with option 2 to fix it for myself, and just manually adjust the offset. 

My advice would be for people to do the same, maybe making it easier by updating the offset when the offset input changes so they don't have to manually grab the DatatableComponent if they didn't want to.

I'm not sure this PR is needed. Thoughts?

Issue #533, opened 2017-02-16
ngx-datatable-column + toggle does not work


eddy-geek:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here
```

**Current behavior**
Setting columns with the ngx-datatable-column directive seems incompatible with toggling columns. Any specific setting (like template) set through the directive will get lost when hiding then showing the column.

**Expected behavior**
Custom settings are retained 

**Reproduction of the problem**
(plunkr link seems outdated, did not work for me)
* replace _column-toggle.ts_ from the included demo with this modified version (makes the **Company** column editable): https://gist.github.com/eddy-geek/9103b001a8c5b1ae563cbb562d728003
* run the demo, toggle **Company** off then on
* observe how it is not editable anymore

**What is the motivation / use case for changing the behavior?**
How else can I toggle columns with custom templates?

**Please tell us about your environment:**
fedora

* **Table version:** 6.0.2

* **Angular version:** 2.3.1

* **Browser:** [all  ]

* **Language:** [TypeScript 2.1.5]

mburger81:
Hi @eddy-geek I think it is the same bug i reported before on this issue?

https://github.com/swimlane/ngx-datatable/issues/529

eddy-geek:
These are separate issues, IIUC your bug is present even when declaring the
columns as json. My issue is when declaring them as html in the template.

On 17 Feb 2017 5:19 pm, "mburger81" <notifications@github.com> wrote:

Hi @eddy-geek <https://github.com/eddy-geek> I think it is the same bug i
reported before on this issue?

#529 <https://github.com/swimlane/ngx-datatable/issues/529>

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/swimlane/ngx-datatable/issues/533#issuecomment-280694664>,
or mute the thread

ajinkyaaa:
you can achieve it using ngFor and ngIf 
<img width="1611" alt="screen shot 2017-07-07 at 10 58 03 am" src="https://user-images.githubusercontent.com/22779150/27963621-9c1fa8d8-6303-11e7-82d0-eba26d842a6e.png">

          <div *ngFor="let col of columns">
            <ngx-datatable-column *ngIf="col.checked == true" prop={{col.prop}}>
              <template let-column="column" let-sort="sortFn" ngx-datatable-header-template>
[...]

PR #551, opened 2017-02-22
Column change detection rework

gerhardboer:
- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe:
Reworked column dataflow to support onPush.

It started with fixing issue #533 where templates were not reapplied when toggling any colum.

**What is the new behavior?**
The templates are reapplied when toggling any column.

**The Bigger issue** 
I could not fix the issue without reworking the way the columns are set. 
Updating the templates called the setter of the columns, and weird things were happening.

There were many internal changes to the objects inside the columns array, which meant Angular can't detect all the changes properly and call the right handlers or hooks.
This can be seen in the demo page, where internal changes 'leaks' out to the column objects set by the developer.
If you put in a simple column, like `{ name: "Name" }`, you'll see that it has more properties when the datatable has processed it internally.

**The Bigger fix** 
I decouple the column objects in every function that changes it, so the original inputted columns remain the same.
By also handling the changes to the columns in the ngOnChange lifycyle hook, Angular has it easier to update the component tree.

- [ ] Yes
- [ ] No
- [x] Maybe

This fix works best when the columns are set to a new array everytime a change is done by the application using the datatable.
This means internally changing the columns array in the application won't update the table.

In my opinion, this is a better way to update the table. It follows the Immutable principle better and i believe this is the recommended way to update components in Angular.

But since there are issues opened where internally changing the columns array are expected to update the table, it would break for those use cases.

**Other information**:
Sorry for committing the release directory, don't really know how to fix that.
If and when this change gets applied, I could create a cleaner branch with the changes so the gitlog remains of high quality.


gerhardboer:
Made a cleaner branch

PR #552, opened 2017-02-22
Template column toggle fix + column handling rework

gerhardboer:
- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe:
Reworked column dataflow to support onPush.

It started with fixing issue #533 where templates were not reapplied when toggling any colum.

**What is the new behavior?**
The templates are reapplied when toggling any column.

**The Bigger issue** 
I could not fix the issue without reworking the way the columns are set. 
Updating the templates called the setter of the columns, and weird things were happening.

There were many internal changes to the objects inside the columns array, which meant Angular can't detect all the changes properly and call the right handlers or hooks.
This can be seen in the demo page, where internal changes 'leaks' out to the column objects set by the developer.
If you put in a simple column, like `{ name: "Name" }`, you'll see that it has more properties when the datatable has processed it internally.

**The Bigger fix** 
I decouple the column objects in every function that changes it, so the original inputted columns remain the same.
By also handling the changes to the columns in the ngOnChange lifycyle hook, Angular has it easier to update the component tree.

- [ ] Yes
- [ ] No
- [x] Maybe

This fix works best when the columns are set to a new array everytime a change is done by the application using the datatable.
This means internally changing the columns array in the application won't update the table.

In my opinion, this is a better way to update the table. It follows the Immutable principle better and i believe this is the recommended way to update components in Angular.

But since there are issues opened where internally changing the columns array are expected to update the table, it would break for those use cases.

**Other information**:

### DIFF
--- a/src/components/datatable.component.ts
+++ b/src/components/datatable.component.ts
@@ -2,17 +2,17 @@ import {
   Component, Input, Output, ElementRef, EventEmitter, ViewChild,
   HostListener, ContentChildren, OnInit, QueryList, AfterViewInit,
   HostBinding, ContentChild, TemplateRef, IterableDiffer,
-  DoCheck, KeyValueDiffers, ViewEncapsulation
+  DoCheck, KeyValueDiffers, ViewEncapsulation, SimpleChange
 } from '@angular/core';
 
 import {
   forceFillColumnWidths, adjustColumnWidths, sortRows, scrollbarWidth,
   setColumnDefaults, throttleable, translateTemplates
 } from '../utils';
-import { ColumnMode, SortType, SelectionType } from '../types';
-import { DataTableBodyComponent } from './body';
-import { DataTableColumnDirective } from './columns';
-import { DatatableRowDetailDirective } from './row-detail';
+import {ColumnMode, SortType, SelectionType} from '../types';
+import {DataTableBodyComponent} from './body';
+import {DataTableColumnDirective} from './columns';
+import {DatatableRowDetailDirective} from './row-detail';
 
 @Component({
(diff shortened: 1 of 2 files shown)


amcdnl:
Thanks for the contribution, however, in Angular4 with the introduction of cold observable streams (https://github.com/angular/angular/issues/13248 ) we can automatically bind to the templates thus eliminating the need for much of this code that also breaks non-immutable objects. Initially, I had switched the whole table to be immutable but we had SOOO many issues with that so I'm very reluctant to reintroduce anything like this.

RC1 dropped yesterday so I'm going to be moving the project to Angular4 as early as next week. We should seem this fixed as a part of that effort. If not, we can certainty revisit this.

gerhardboer:
I saw the commit that added it and the revert, so i knew of the issues. It's why I tried to get some more real world examples before adding a pull-request, since it seemed to work on the demo pages.

But if the fix comes from Angular, that would be the best way 👍

I still think being able to put the datable to onPush would be the best for overall application performance, since it would not interfere with any changes to other inputs in the application.
This is quite noticable with big tables and animations.

Also with the new ViewEngine, there will be a performance decrease in the templates.
The recommendation to keep the application fast is to put the components to OnPush, so supporting it might get quite important.
src: https://docs.google.com/document/d/195L4WaDSoI_kkW094LlShH6gT3B7K1GZpSBnnLkQR-g/pub

amcdnl:
@gerhardboer - so little known fact, if u declare your parent component as `OnPush` it trickles down until otherwise stated ( which I do not in the table ).

PR #553, opened 2017-02-22
fix: show first page after sorting

### LINKED ISSUE #270: Server side paging prolem
Hi
Please do the steps to get the error:
1- Goto: http://swimlane.github.io/angular2-data-table/
2- Left bar Paging > Server-side
3- Click on page number 4 or 5 
4- Now, try to sort on Company

Table be empty, until you click on page 1 .

deeg:
BREAKING CHANGES: For client side pagination with no vertical scroll, if you sort a column after navigating to any page but the first, you will be brought back to the first page rather than seeing the sorted results on the page you were currently on.

If you were using a vertical scroll, the sorting was already automatically pushing you back to the first page. This should be done for any sort event, so now regardless of your configuration, on sort you will be brought back to the first page.

Closes #270

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

**Using Vertical Scroll**

When sorting by a column, you are always brought back to the first page

**Client side sorting without vertical scroll**

When you sort you see the sorted results, on the page you are currently on. So if you are on page three and you sort by name alpha, you might be looking at the letter C, instead of going to page one and seeing letter A.

**Server side sorting without vertical scroll**

When you sort from a page which is not the first, you remain on that page and see no data.

**What is the new behavior?**

You always get brought back to the first page regardless of configuration when you are sorting.

- [X] Yes
- [ ] No

Described in **Client side sorting without vertical scroll** above.

### DIFF
--- a/src/components/datatable.component.spec.ts
+++ b/src/components/datatable.component.spec.ts
@@ -5,7 +5,7 @@ describe('Datatable component', () => {
 
   beforeEach(() => {
     TestBed.configureTestingModule({
-      imports: [ NgxDatatableModule ]
+      imports: [NgxDatatableModule]
     });
   });
 
@@ -18,9 +18,9 @@ describe('Datatable component', () => {
     it('should return a new array', () => {
       let fixture = TestBed.createComponent(DatatableComponent);
       let initialRows = [
-        { id: 1 },
-        { id: 2 },
-        { id: 3 }
+        {id: 1},
+        {id: 2},
+        {id: 3}
       ];
 
       let columns = [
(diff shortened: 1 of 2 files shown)


yoAlignment:
same problem,but I want to fix this in angular@2.3.1.so that ngx-dataTable@7.0.0 will not allowed to install,what should I do.

matte00:
Is there a regression? When sorting by a column, you are always brought back to the first page. 
I check it on ver 10.2.3 and in your demo page example: [http://swimlane.github.io/ngx-datatable/#client-paging](url) or [http://swimlane.github.io/ngx-datatable/#server-paging](url)

matte00:
Is there an option to keep the current page when sorting?

PR #565, opened 2017-02-27
demo(filtering): go to first page on filter change

### LINKED ISSUE #375: when using pagination and switching to page>1, and then filtering the rows, no results are shown
```
[X] bug 

**Current behavior**
When using pagination, and switching to page>1, and then filtering data, the filtered data is not shown (unless there is enough data to populate the pages)


**Expected behavior**
I think the table should switch back to page 1 on data change or via a manual method call.


see plunker :
https://plnkr.co/edit/cQKNI9cDMkLCJrcWtdcK

click "page 2" and then click "filter data". the data is gone although there is one record.

I need to combine pagination and data filtering on the table.

latest

* **Table version:** 0.7.x
latest

* **Angular version:** 2.0.x
latest

Browser : Chrome

### LINKED ISSUE #515: Bug with filtering data

### PR DESCRIPTION
**Pull request by deeg**
This fixes an issue with the demo where the table did not navigate back to the first page after updating the filter.
Closes #515
Closes #549 
Closes #375

There is a decision to be made here, and that is how we want the user to programmatically change which table the page is on. Right now this PR fixes the filtering demo, by programmatically going back to the first page whenever there is a filter change.

Right now I believe this is the only way to do it, due to the nature of Angular's input change detection. Ideally you would also be able to change the offset by changing the value of the [offset] input of the table. Right now that doesn't work because it is a primitive. I think in order to make that work we would have to change offset to something like {value: 0} instead of just a number.

[See this commit which demonstrates how we could change offset to programmatically change it by updating the input](https://github.com/deeg/angular2-data-table/commit/aeab1774eb1d7097f3c739c590a6842bfe99374d)

If you were to pull down the branch and run the filtering demo:

* navigate to page 4
* type f in the filter
* and the console would look like:
```
changing offset
0
changing test offset
Object {value: 0}
changing offset
3
changing test offset
Object {value: 0}
```

I believe due to the way angular works, the `offset setter` doesn't fire an additional time like the `testOffset setter` since it is a primitive. If I'm mistaken about that, let me know! Maybe there is an easy way to get the primitive change to fire the setter.

It might not be worth making that change (offset being an object), and just making this the default way of programmatically changing the page, or possibly add in another method to the table like `recalculate`.

There is also a decision to be made on if we want the table to protect against this somehow or let the user be in full control of it. This solution proposes it is up to the user to decide how to handle it, rather than the table trying to decide which page to navigate to depending on the rows array.

@amcdnl, thoughts?

- Demo changes which now shows how to programmatically change a page
- [ ] Yes
- [X] No

### DIFF
--- a/demo/basic/filter.ts
+++ b/demo/basic/filter.ts
@@ -1,4 +1,5 @@
-import { Component } from '@angular/core';
+import {Component, ViewChild} from '@angular/core';
+import {DatatableComponent} from '../../src/components/datatable.component';
 
 @Component({
   selector: 'filter-demo',
@@ -12,6 +13,7 @@ import { Component } from '@angular/core';
         (keyup)='updateFilter($event)'
       />
       <ngx-datatable
+        #table
         class='material'
         [columns]="columns"
         [columnMode]="'force'"
@@ -35,7 +37,8 @@ export class FilterBarComponent {
     { name: 'Company' },
     { name: 'Gender' }
   ];
-  
+  @ViewChild(DatatableComponent) table: DatatableComponent;
+
   constructor() {
     this.fetch((data) => {
       // cache our list
@@ -67,6 +70,8 @@ export class FilterBarComponent {
 
     // update the rows
     this.rows = temp;
+    // Whenever the filter changes, always go back to the first page
+    this.table.offset = 0;
   }
 
 }


amcdnl:
Thanks as always!

PR #570, opened 2017-02-27
Datatable items don't return to original order when columns return to 'unsorted' state.

tw3:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

1) Go to the demo page:
http://swimlane.github.io/ngx-datatable/#
2) Navigate to Sorting > Client-side in the left navigation
3) Note the original order of items on the datatable, e.g. "Johnson, Johnson" is shown first
4) Sort the datatable by the Company column, items are sorted ascending
5) Sort the datatable again by the Company column, items are sorted descending
6) Sort the datatable yet again by the Company column, Company column is back to unsorted state yet the order remains descending

**What is the new behavior?**
When returning to the unsorted state for all columns (step 6 above) the original order of items in the datatable is restored.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -298,7 +298,7 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
     let offset = this.indexes.first / this.pageSize;
 
     if(direction === 'up') {
-      offset = Math.ceil(offset);
+      offset = Math.floor(offset);
     } else if(direction === 'down') {
       offset = Math.ceil(offset);
     }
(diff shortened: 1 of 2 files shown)


wyers:
Disclosure, I work with @tw3.

In my opinion, this would only effect client-side sort. I feel like 'unsorted' should be equivalent to 'reflect my source array'. If there is some external factor that is continually editing that source array, I think those changes would be expected. 

In most scenarios, my users want to quickly sort by name, then go back to the original view (sent once from the server) that is sorted by date last modified. They've expressed confusion/frustration at having to manually resort (if that has any influence).

Also, it's a great tool. Thanks for making it available!

Issue #610, opened 2017-03-21
Do not mutate objects


shlomiassaf:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[x] feature request
```

**Current behavior**
The table mutate the objects / raws with an `$$index` property. [code](https://github.com/swimlane/ngx-datatable/blob/master/src/components/body/body.component.ts#L328)

**Expected behavior**
Do not mutate objects.

**What is the motivation / use case for changing the behavior?**
It is considered a bad practice to mutate objects sent by the user.
With today's modern language constructs it can be easily avoided.

The issues vary, here is a summery:
  1. Impossible to handle immutable objects
  2. Unintentional reflection of the `$$index` property
  3. Usage conflict (the user or other 3rd party lib's might be using `$$index`

The user is not aware that this happens and this might create serious implications with the current implementation.

For example:

A user might use the **ImmutableJS** library or use **Object.freeze/seal** on the objects sent as rows, this will cause an exception by the library that can't handle these objects. (ImmutableJS is a quite common library)

If an object is sent to the table and then used to create an HTTP request the `$$index` property will be sent along with the request. This is unintentional reflection.

### There are several solution to solve this, this is my preferred solution:

  - Use `WeakMap` to store the index of every row, rows are automatically removed from the `WeakMap` instance  when removed from the table (and no longer referenced by the user). Each table instance should have a `WeakMap` instance. 
It is possible to remove the row from the weekmap when removed from the table.

```ts
  // somewhere at construction: 
  this.idxMap = new WeakMap() 

  // set index:
  this.idxMap.set(row, rowIndex);


 // use this.idxMap.get to fetch the index.
```

This solution can't create conflicts, it will not reflect on object serialization and it will not mutate the object.

##### Other possible solutions:

  - Use a `Symbol` instead of a property, symbol are guaranteed not to conflict with any other library.

```ts
  // somewhere at module's top: 
  IndexSymbol = Symbol('ngx-datatable RowIndex');

  // set:
  row[IndexSymbol] = rowIndex;

 // get
 row[IndexSymbol]
```
> This solve any conflict and will not reflect in serialization **but it will throw on immutable objects.**

  - Assign `$$index` with a property descriptor so the value is not **enumerable**.
```ts
  Object.defineProperty(row, '$$index', { value: rowIndex });
```
> This does not solve usage conflicts  and mutates the object, not good.

amcdnl:
[...]

PR #623, opened 2017-03-22
Disable row mutation

shlomiassaf:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

See #610

**What is the new behavior?**
See #610

- [X] Yes
- [ ] No

Rows are no longer mutated with `$$index` or `$$expanded`

Since not there, all custom templates that used these properties will no longer work.
To facilitate this, the `row` object sent to **all** templates is now of type `RowMeta` which is a wrapper around the original row object sent by the user.

```ts
export interface RowMeta {
  row: any;
  rowIndex: number;
  expanded?: number;
}
```

> NOTE: Future refactoring might add generic (`RowMeta<T>`) which leads to having `DatatableComponent<T>` at the top. This can wait till TS 2.3 is out and widely spread so we can assign a default to T and not bother our users with a breaking change.

Since the metadata is important for user logic, **all events** except **select** event now has `row` property as `RowMeta`. This means we can now easily extend this datatype to add more info as needed.

The select event send's the user supplied row instances.

The reason for this is that the `selected` **@Input** property excepts raw row instances from the user, if a user supplies this array the selector will have to traverse the whole array and find the index for each selected item, so I skipped that for the time being.
The drawback is that you can't get the selected indicies.

## Implementation:
Turns out that `WeakMap` is not suitable for this task.
The logic implemented throughout the library is using iteration over an array, this means that in a row array iteration in order to find a `RowMeta` there's going to Map lookup for every iteration.
Map is O(1) but still not as fast as an array index lookup, also a lot of browsers might use a polyfil that is far from O(1).

Instead, I kept the array logic and passed an array throughout the library which made the change easy to handle. The array is now of `RowMeta` instead of the user's raw rows.  
Now the only change is to access the `RowMeta#row` inside the receiving functions.

The array is lazy populated, it starts as an empty array with the length N (where N is the rows length) and get's populated by demand in `DataTableBodyComponent#updateRows()`.  
It fit's perfectly with `RowHeightCache` since we send `RowMeta` array, if the value is not in a specific index it's definitely not expanded.

### DIFF
--- a/demo/basic/row-detail.ts
+++ b/demo/basic/row-detail.ts
@@ -1,4 +1,5 @@
 import { Component, ViewEncapsulation, ViewChild } from '@angular/core';
+import { RowMeta } from "../../src/types/row-meta";
 
 @Component({
   selector: 'row-details-demo',
@@ -31,7 +32,7 @@ import { Component, ViewEncapsulation, ViewChild } from '@angular/core';
           <ng-template let-row="row" ngx-datatable-row-detail-template>
             <div style="padding-left:35px;">
               <div><strong>Address</strong></div>
-              <div>{{row.address.city}}, {{row.address.state}}</div>
+              <div>{{row.row.address.city}}, {{row.row.address.state}}</div>
             </div>
           </ng-template>
         </ngx-datatable-row-detail>
@@ -46,8 +47,8 @@ import { Component, ViewEncapsulation, ViewChild } from '@angular/core';
           <ng-template let-row="row" ngx-datatable-cell-template>
             <a
               href="#"
-              [class.icon-right]="!row.$$expanded"
-              [class.icon-down]="row.$$expanded"
+              [class.icon-right]="!row.expanded"
+              [class.icon-down]="row.expanded"
               title="Expand/Collapse Row"
               (click)="toggleExpandRow(row)">
             </a>
@@ -55,12 +56,12 @@ import { Component, ViewEncapsulation, ViewChild } from '@angular/core';
         </ngx-datatable-column>
         <ngx-datatable-column name="Index" width="80">
           <ng-template let-row="row" ngx-datatable-cell-template>
-            <strong>{{row.$$index}}</strong>
+            <strong>{{row.rowIndex}}</strong>
           </ng-template>
         </ngx-datatable-column>
         <ngx-datatable-column name="Exapanded" width="80">
           <ng-template let-row="row" ngx-datatable-cell-template>
-            <strong>{{row.$$expanded === 1}}</strong>
+            <strong>{{row.rowIndex === 1}}</strong>
           </ng-template>
         </ngx-datatable-column>
         <ngx-datatable-column name="Name" width="200">
@@ -70,7 +71,7 @@ import { Component, ViewEncapsulation, ViewChild } from '@angular/core';
         </ngx-datatable-column>
         <ngx-datatable-column name="Gender" width="300">
           <ng-template let-row="row" let-value="value" ngx-datatable-cell-template>
-            <i [innerHTML]="row['name']"></i> and <i>{{value}}</i>
+            <i [innerHTML]="row.row['name']"></i> and <i>{{value}}</i>
           </ng-template>
         </ngx-datatable-column>
         <ngx-datatable-column name="Age" ></ngx-datatable-column>
@@ -112,7 +113,7 @@ export class RowDetailsComponent {
     req.send();
   }
 
-  toggleExpandRow(row) {
+  toggleExpandRow(row: RowMeta) {
     console.log('Toggled Expand Row!', row);
     this.table.rowDetail.toggleExpandRow(row);
   }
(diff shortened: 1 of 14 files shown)


shlomiassaf:
@amcdnl I have change this code:
```ts
      if(rowMeta) {
        rowMeta.rowIndex = rowIndex;
        temp[idx++] = rowMeta;
      }

      // idx++;
      rowIndex++;
```

Removed the set of instructions and move into one place.

I don't see any reason why we need to bump the index if there is not row, it will leave an empty one.

In general I think `rowMeta` will always be populated... but I don't know so I left the other instruction

demo/basic/row-detail.ts
```diff
@@ -31,7 +32,7 @@ import { Component, ViewEncapsulation, ViewChild } from '@angular/core';
           <ng-template let-row="row" ngx-datatable-row-detail-template>
             <div style="padding-left:35px;">
               <div><strong>Address</strong></div>
-              <div>{{row.address.city}}, {{row.address.state}}</div>
+              <div>{{row.row.address.city}}, {{row.row.address.state}}</div>
```
amcdnl: Dont love the `row.row` syntax. Any ideas on better idea?

amcdnl:
@shlomiassaf - First off, amazing work! I'll be merging this asap! Thanks for your contribution!

amcdnl:
Merged into master but have not released yet, check out the changelog notes and let me know if I missed anything:

- https://github.com/swimlane/ngx-datatable/commit/419132bdaccac1fe28a65446e4b9726c682f80c1#diff-343723a0cb380f664aedf0dfed5c1c2fR8

Also, instead of doing `row.row`, I made `row` emit the normal row like it was before and added a new `meta` attribute, see: https://github.com/swimlane/ngx-datatable/commit/419132bdaccac1fe28a65446e4b9726c682f80c1#diff-a08ce11154abcc80d58a8072a06283ebR30

shlomiassaf:
@amcdnl 👍 

I think the changelog covers most of it, the only missing piece is getting the row but index.  
This is a design block I had, the implementation allows getting row instances in the **select** `@Input` which makes it difficult to fire out indicies since we use the object ref as an identifier and not the index in the array.

If the user of the table can't supply a pre selected rows we can provide the selected indicies AND selected rows.
If the user supplies selected rows - we can provide selected rows output
If the user supplies selected indicies - we can provide selected indicies output.

All in all this is limiting but not blocking since the user can find the index himself, if he has a row ref.

About the diff, no problem. I agree that `row.row` is a poor API. I just couldn't make the decision to add another local template variable myself.

Note that `meta` will output an object that has the `row` property as well, so this should be documented... row is syntax sugar or a shortcut... but if you already declare `meta` as template var no reason to declare row. Probably should alter on of the demo examples to show that.

amcdnl:
Ya, I wondered about the selections, is selection event emitting the rowmeta now? 

I can't see a user case where you only have the indexes you want to select vs the actual row object. If it works where users pass the rows (not meta) now, I think its suitable.

Ya, I think most ppl won't want the meta object though.

shlomiassaf:
The selections will emit `row`, not reason to emit meta there.

amcdnl:
Why did you change contextmenu to emit meta? We should probably keep that consistent ....

shlomiassaf:
My thinking was that it's important for the handler to know the internal state.

expanded might be useful, index as well...

amcdnl:
Why is `expanded` a number? Shouldn't it be a boolean?

I tend to disagree, I think this is an internal meta object. The expanded is a separate method the user should use. The 'real' index in the row is also not important to a user IMO.

shlomiassaf:
`expended` is a number because it is used as a number throughout the code.... 1, 0 and some bitwise on them.

It was `any` and I didn't want to start changing the logic of expended so I had to do `number` to satisfy typescript.

The `RowMeta` is indeed an internal object but as I said before, it is information you want to expose in cases where the user changes behaviour or needs to get information about the row.

Not including it will reduce the extensibility one can do from the outside, which is not what the library aims for... light, small and extensible.

Now, exposing `RowMeta` also plays along with a plugin architecture where you can extend the **type** (`RowMeta`) from the outside and add metadata to it.

Say you want to count the times the context menu was opened, per row...

It's ok to expose a wrapped object in some scenarios.

You can revert this change and return the row only, it is totally you'r call but you will face many issue where users are missing that piece of information and can't do their crazy things... i'm 100% sure on that.

shlomiassaf:
As a side note, I think we should open an issue to do some refactoring on the code, virtual refactoring... add a lot of missing type information throughout.

A lot of events fire `any`...
a lot of internal objects have most of their public properties as `any`

This made it hard for me to understand the code, as a newcomer... it also more prone to bugs since TS can't help

amcdnl:
I'm going just return the row for this release to minimize breaking changes. If it comes up from someone, we can eval and add it then. Thanks again for your efforts!

amcdnl:
Sadly, I had to revert this commit. There was a lot of issues with the row detail that is gonna take quite a bit of thought. I made a branch with my previous changes called `mutations`. Lets work from there to get this totally working.

shlomiassaf:
@amcdnl That's unfortunate.

I didn't get any issues from the unit tests and didn't spot any on the demo (but that was eye scan, not logic tests)

I guess these are issues you find in your internal project implementing the table...

Anyway, if I know them I can try and help, if they come up from internal projects I can't do much

PR #662, opened 2017-04-06
Add the ability to deactivate the paging on scroll

evermax:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

When using the virtual scroll, the page is changing.

**What is the new behavior?**
Add an option `scrollPaging` that is defaulted to `true` to avoid any breaking changes.
If set to `false` the data table will ignore `page` event from the body, leaving the user to only use the footer's paging.

- [ ] Yes
- [x] No

**Other information**:
I branched off of the 6.3.0 release commit as we are using 6.3.0. Do you plan to maintain 6.x.x versions?

### DIFF
--- a/src/components/datatable.component.ts
+++ b/src/components/datatable.component.ts
@@ -214,6 +214,15 @@ export class DatatableComponent implements OnInit, AfterViewInit, DoCheck {
   @Input() externalPaging: boolean = false;
 
   /**
+   * If the table should use pagination on scroll
+   * otherwise it only uses the pagination in the footer
+   * 
+   * @type {boolean}
+   * @memberOf DatatableComponent
+   */
+  @Input() scrollPaging: boolean = true;
+
+  /**
    * If the table should use external sorting or
    * the built-in basic sorting.
    *
@@ -800,14 +809,16 @@ export class DatatableComponent implements OnInit, AfterViewInit, DoCheck {
    * @memberOf DatatableComponent
    */
   onBodyPage({offset}: any): void {
-    this.offset = offset;
+    if (this.scrollPaging) {
+      this.offset = offset;
 
-    this.page.emit({
-      count: this.count,
-      pageSize: this.pageSize,
-      limit: this.limit,
-      offset: this.offset
-    });
+      this.page.emit({
+        count: this.count,
+        pageSize: this.pageSize,
+        limit: this.limit,
+        offset: this.offset
+      });
+    }
   }
 
   /**


evermax:
Accessing the datatable using `@ViewChild` and overriding the `onBodyPage` function did the trick.
I still think something around those lines would be nice, but I am going to close this for now.

RaedsLab:
Is there currently a way to have scrolling `[scrollbarV] ` with no paging?

As in scroll in the same page and have fixed footer and header.

wizarrc:
https://github.com/swimlane/ngx-datatable/blob/master/src/components/body/body.component.ts#L332-L347
```
  /**
   * Updates the page given a direction.
   */
  updatePage(direction: string): void {
    let offset = this.indexes.first / this.pageSize;

    if (direction === 'up') {
      offset = Math.ceil(offset);
    } else if (direction === 'down') {
      offset = Math.ceil(offset);
    }

    if (direction !== undefined && !isNaN(offset)) {
      this.page.emit({ offset });
    }
  }
```
@RaedsLab If you look there you can see that it's always fired on every `scroll` event. A simple variable could detect when `offset` is changed and only fire then. Would that acceptable?

wizarrc:
@RaedsLab Can you set the table height in css, and turn off virtual scrolling? A side effect would be all items would be rendered, even outside of the view port.

RaedsLab:
@wizarrc  Thank you ... here is my setup now:

```html 
<ngx-datatable    style="height: 400px;" 
				   class="material orbit" 
				   [columns]="columns"
				   [rows]="rows" 
				   [columnMode]="'force'" 
				   [headerHeight]="60"
				   [footerHeight]="50" 
				   [rowHeight]="50" 
				   [scrollbarV]="false"
			           [scrollbarH]="false" 
....
```

I can scroll using keys up and down , but I don't get a scroll bar.

wizarrc:
@RaedsLab hmm, do you have a plunkr? It would be easier for me to see what your issue is.

RaedsLab:
http://plnkr.co/edit/pZ30I4OjOHHzFT5slpqu?p=preview

wizarrc:
@RaedsLab Thanks for the demo. Try adding the following to your style where `height` is set to `400px`, and header and footer heights are 110px:

```
.ngx-datatable .datatable-body {
  overflow-y: scroll;
  height: 290px !important; // 400px - 60px - 50px (header and footer)
}
```
notice how `!important` is used because the `height` property is set to `auto` via inline style.

PR #685, opened 2017-04-17
refactor: enable no implicit any check for library

clydin:
- [X] Refactoring (no functional changes, no api changes)

implicit any check was disabled.  The public API contained several uses of an implicit any which can cause errors in projects using the library.

**What is the new behavior?**

Implicit any check is enabled for the library code and all instances are corrected.

- [X] No

**Other information**:


amcdnl:
Thanks for your contribution!

Issue #115, opened 2016-09-12
Feature Request: Summary Row


Tempus35:
Another request:

The ability to have a row that is either in the footer or just pinned above the footer and after the body that can be hold templates like a normal row. The use case here is to allow a summary of numerical data in a column that will always be on the screen no matter the page.

Tempus35:
A visual of this can be found here: [http://imgur.com/a/2VziI](http://imgur.com/a/2VziI).

amcdnl:
Like this: http://demos.telerik.com/kendo-ui/grid/aggregates

Tempus35:
Basically, would be nice of I could use a template ref in the cells in the summary row

phillipadamson:
Hey all, I just wanted to check to see if any progress has been made on this, as this is something that would really add value to this great project.

Tempus35:
#694

amcdnl:
Resolved by custom footer in 8.2.0

PR #694, opened 2017-04-19
Basic Summary Row

Tempus35:
- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

No row to summarize the data in a column #115

**What is the new behavior?**
Adds a summary row that allows you to add summary information for each column, including a label.

- [ ] Yes
- [X] No

**Other information**:
This introduces a basic summary row. At the moment it mimics the body row in many ways though it lacks a template ref, since it is likely you would want a template in the summary that is not the same as the body cells. I will look into adding a summaryCellTemplate later. 

You can send in two things to the datatable for the summary row:
**summaryRow** and **summaryRowHeight**
The row is an object that can be in two formats:

**{
  name: 'John'
}**

or 

**{
  name: { label: 'First Name', value: 'John'}
}**

If it finds the label it will add it and bold it before the value otherwise it just uses the value.


amcdnl:
First good cut! Few things...

1) If you turn on scrollbarV, the footer doesn't show. I fixed this in a commit
2) The summary row should mirror the row/column api. Right now it kinda feels weird as a totally different API.
3) You define `Age` column but it doesn't show up.
4) It needs to have a declarative api like columns.

amcdnl:
I actually think a custom footer is a better approach for this. I even made a demo in the demos page for it ;O.

Custom footer is in 8.2.0

Issue #703, opened 2017-04-23
Tables with cell details, virtual scroll and dynamic cell detail height do not work


Szabadember:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
Table row details are shown behind each other instead of under each other. Also the layout gets completely screwed up, it is impossible to scroll anymore. It is impossible to open or close row details.

**Expected behavior**
Rows should be displayed under each other and the user should be able to open and close the detail panel.

**Reproduction of the problem**
Here is a plunker for easier reproduction:
http://plnkr.co/edit/C0bWPAgQCJR90auNGA25
I also attached the source of a typescript component which can be tested using the ngx-datatable demo app by replacing the contents of dynamic-height.component.ts.

Steps to reproduce:

1. Open all the row details starting from the last row
2. Scroll down to make the first row get completely out of the view (the second row should be on top)
3. Close the row details starting from the last row

**What is the motivation / use case for changing the behavior?**
I want to build a table where each row can have it's own row detail view with different height

**Please tell us about your environment:**
Operating system: OS X 10.12.4
IDE: VS Code

* **Table version:** 0.8.x
Table version: 8.0.0

* **Angular version:** 2.0.x
Angular version: 4.0.3

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
Chrome, Safari, Firefox

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
TypeScript 2.3.0

```
import { Component, ViewChild } from '@angular/core';

@Component({
  selector: 'dynamic-height-demo',
  template: `
    <div>
      <h3>
        Dynamic Height w/ Virtual Scrolling
        <small>
          <a href="https://github.com/swimlane/ngx-datatable/blob/master/demo/basic/dynamic-height.component.ts" target="_blank">
            Source
          </a>
        </small>
      </h3>
      <ngx-datatable
        #myTable
        class="material expandable gk-toplevel-container"
        [columnMode]="'force'"
        [headerHeight]="40"
        [footerHeight]="50"
        [rowHeight]="40"
        [scrollbarV]="40"
        [rows]="yearlySums">
[...]

PR #704, opened 2017-04-23
Fixed dynamic row detail height issue

Szabadember:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When the row height cache gets refreshed, getDetailRowHeight gets called without passing the row or index parameters. After the row details are expanded, this causes the cache to fill up with wrong heights.

**What is the new behavior?**
Changed initCache so that it uses the same logic for determining detailRowHeight as for rowHeight. It checks whether the passed detailRowHeight is a function or a number and acts accordingly (calls the function with the current row and adds it's return value to the row height or if it is a number then it adds it directly to the row height). This should guarantee that the cache is filled up with the correct heights.

- [ ] Yes
- [x] No

**Other information**:
This PR fixes the following issue:
https://github.com/swimlane/ngx-datatable/issues/703


amcdnl:
In the future, please don't commit build files. Thanks for your contribution!

Issue #138, opened 2016-09-21
Server pagination not working


SnakeB:
Pagination and Sorting should be handled through the server. To load my data from the backend I´m using the following method:

```
this.loadData('sortColumn', 'ASC', this.currentPage, this.options.limit)
        .subscribe(ratings => {
        this.rows = [];
        this.rows.push(...ratings.ratings);
        this.endRow = this.options.limit * (this.options.offset + 1);
        this.maxRows = ratings.recordsTotal;
        if (this.maxRows < this.options.limit) {
          this.endRow = this.maxRows;
        } else if (this.maxRows < this.endRow) {
          this.endRow = this.maxRows;
        }
        this.options.count = this.maxRows;
  }, error => this.errorMessage = <any>error);}
```

this.rows is the variable I´m binding to the data-table-template:

`<datatable class="rating_table" [rows]='rows' [options]='options' (onRowClick)='onRowClick($event)' (onPageChange)='onPage($event)'>`

My problem is when I´m loading the first page the data binding ( this.rows.push(...ratings.ratings))  works and in the table my ratings are shown, but when I´m loading the second page the new ratings are saved into the this.rows object but the table is not refreshing the ratings from page one are still visible.

amcdnl:
Please make a demo in plunkr.

juracy:
Same problem here, it looks like angular2-data-table is looking for options.offset + 1 in rows, instead of getting a completed new "buffer" starting from 0...

amcdnl:
Offset + 1 is because offset is 0 based, for paging there is no 0. :)

SnakeB:
I will prepare and post an example tomorrow

SnakeB:
I prepared a small example to show the problem: 
https://plnkr.co/edit/jTf0ospsvh5TzhOQtdQp?p=preview
I´m setting new rows every time the user switches from one page to another:
```
this.rows = [];
    if(pageNumber === 0){
      this.rows.push(...this.page1Data);
    }else{
      this.rows.push(...this.page2Data);
    }
    this.options.count = this.page1Data.length + this.page2Data.length;
    console.log(this.rows);`
```
if you look at the console the rows object is updated with the new data but the page is not refreshing. I hope you can help me.

SnakeB:
Could you please help me with my problem

robertbachmann:
I had a similar problem. The trick is to update the rows Array in a different way. See 
[...]

PR #714, opened 2017-04-25
Server pagination fix

dcesiel:
This change addresses #138.  You can try it using `npm i @dcesiel/ngx-datatable`

TODO:
1. Add documentation for change.
2. Modify server paging demo for change.

- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently ngx-datatable calculates a non-zero first index in BodyComponent.updateIndexes() when server pagination is selected.  This becomes an issue when using standard pagination such as [spring-data pagination](http://ankushs92.github.io/tutorial/2016/05/03/pagination-with-spring-boot.html) where the first value in the results array will be the value at the specified offset in the request.

**What is the new behavior?**
Because the server is handles grabbing the page subset for us, the first index calculated by BodyComponent.updateIndexes() should always be 0 and the last should be 0 + pageSize.

- [ x ] Yes
- [ ] No

The externalPaging config flag has been deprecated in favor of serverPaging which behaves the same way except for the index calculation change.

**Other information**:

### DIFF
--- a/demo/paging/mock-server-results-service.ts
+++ b/demo/paging/mock-server-results-service.ts
@@ -1,0 +1,43 @@
+import {Injectable} from "@angular/core";
+import {Observable} from "rxjs";
+import {PagedData} from "./model/paged-data";
+import {CorporateEmployee} from "./model/corporate-employee";
+import {Page} from "./model/page";
+var companyData = require('../../assets/data/company.json');
+
+/**
+ * A server used to mock a paged data result from a server
+ */
+@Injectable()
+export class MockServerResultsService {
+
+    /**
+     * A method that mocks a paged server response
+     * @param page The selected page
+     * @returns {any} An observable containing the employee data
+     */
+    public getResults(page: Page): Observable<PagedData<CorporateEmployee>> {
+        return Observable.of(companyData).map(data => this.getPagedData(page));
+    }
+
+    /**
+     * Package companyData into a PagedData object based on the selected Page
+     * @param page The page data used to get the selected data from companyData
+     * @returns {PagedData<CorporateEmployee>} An array of the selected data and page
+     */
+    private getPagedData(page: Page): PagedData<CorporateEmployee> {
+        let pagedData = new PagedData<CorporateEmployee>();
+        page.totalElements = companyData.length;
+        page.totalPages = page.totalElements / page.size;
+        let start = page.pageNumber * page.size;
+        let end = Math.min((start + page.size), page.totalElements);
+        for (let i = start; i < end; i++){
+            let jsonObj = companyData[i];
+            let employee = new CorporateEmployee(jsonObj.name, jsonObj.gender, jsonObj.company, jsonObj.age);
+            pagedData.data.push(employee);
+        }
+        pagedData.page = page;
+        return pagedData;
+    }
+
+}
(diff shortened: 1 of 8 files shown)


amcdnl:
I'm ok with a breaking change, I think providing both options is confusing ( it confused me at first even ).

Can you update the demo page for this change too please?

efriandika:
It works for me.. 
Thanks @dcesiel 💃

dcesiel:
@amcdnl I'll fix the demo as soon as I get time here. 

I'm not sure I agree with your decision on the breaking change. One of the major issues I've always had with javascript/frontend libraries is their disregard for breaking changes, and this change will break all existing uses of external paging. I think we should try to do a better job explaining why a new config boolean was created rather than break existing implementations, but you are the package maintainer and I will go with your decision.

dcesiel:
The most recent commit contains a demo and removes the extra serverPaging flag in favor of the existing externalPaging flag.

rafaelss95:
Hello, I've installed this package and after hours trying to make the externalPaging I've found this thread..

So, I've installed it using `npm i @dcesiel/ngx-datatable`, but I'm getting the same behavior:

The `rows` variable isn't being updated for some reason. It only displays correct on page load, in other words, for the first page.

I put a `console.log(this.rows)` and it displays correct, but the view isn't being updated. How can I solve this?

Here's my actual code:

` <ngx-datatable class="material"
  columnMode="force" 
  [count]="count" 
  [cssClasses]="cssClasses" 
  [externalPaging]="true" 
  [externalSorting]="true" 
  [footerHeight]="50" 
  [headerHeight]="50" 
  [limit]="20" 
  [offset]="0" 
  [rows]="items" 
  rowHeight="auto" 
  (page)="handlePageChange($event)"
  (sort)="handleSortChange($event)">
  </ngx-datatable>`

  ```
public fetchData = (params: { [key: string]: any } = {}): void => {
    this.myService.getAll(params)
      .finally(() => this.loading = false)
      .subscribe(
        (data: any) => {
          this.count = data['meta']['total'];
          this.rows = data['myKey'];
          console.log(this.rows); // It's correct
          // this.rows = this.rows.slice(); -> Attempt
        },
        (err: any) => {
          this.handleError(err);
        }
      );
  }

  public handlePageChange = (event: { [key: string]: number }): void => {
    const {limit: by_limit, offset: by_offset} = event;
    this.fetchData({by_limit, by_offset});
  }
```

efriandika:
Hy @rafaelss95 .. I installed this package few days ago, It works properly.
Please ensure your backend pagination logic is work properly as well

rafaelss95:
Hello @efriandika, thanks for your comment. Yep, it's okay, I just tested it many times before post it here :(

dcesiel:
@rafaelss95 Change `[externalPaging]="true"` to `[serverPaging]="true"`.  I added the `externalPaging` boolean to that package to prevent a breaking change, although it won't go in the final version of ngx-datatable which will continue using `serverPaging`.

rafaelss95:
@dcesiel Man, thank you so much. Finally it's working :+1:

efriandika:
Hopefully @amcdnl merge this pull request as soon as possible.. :D

amcdnl:
Merged. Just needed to test :)

amcdnl:
Merged and released in 9.0! I also added virtual server-side paging! :)

efriandika:
Thanks all of you.. 😁👍

PR #729, opened 2017-05-02
Renamed icons to fix conflicts with other libraries.

gearsandcode:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

icon classes begin with icon- and font-family is set to !important. This is causing it to override the icon font-family in other libraries.

**What is the new behavior?**
Renamed icon-* to datatable-icon-*

- [x] Yes
- [] No

If someone has specified their icon classes using [cssClasses] and they reference the old names, they will no longer display.

**Other information**:


Tempus35:
Nice, my only concern are these lines:
`<div class="icon datatable-icon-filter"></div>`
The icon should be datatable-icon right, to prevent issues?

gearsandcode:
It could likely just be removed as datatable-icon-filter is specific enough. I don't see that it was used before as the wildcard selectors has the - suffix on them.

amcdnl:
Awesome! thanks for your contrib!

Issue #611, opened 2017-03-21
detect vertical scrolling to the end of the table


xhe:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[x ] feature request
[ ] support request => Please do not submit support request here
```

**Current behavior**
Hello, may I ask how to detect scrolling to the end of the table? we need this feature because we have infinity scrolling. I think I may use "scroll" event, but that require me to do calculation. If  there is an event dispatched when scroller is approaching the end, that will be great.

Maybe we already have this event?

amcdnl:
So you just want a event to fire when you hit the end of the scroll?

xhe:
yes, is this doable?

xhe:
Hello,

I noticed if I am rending more than 20 columns, the dataGrid is very slow.

generating data is very fast, but rendering the table takes upto 10 seconds.

Anything wrong with the code?

I am creating a component wrapper using ngx-datatable, and then passing
rows/columns, not seeing anything wrong.

Please help to check.

Thanks

Frank

amcdnl:
Are you running angular in prod mode? Dev mode its kinda slow due to the introspection they do.

xhe:
Oh, a good hints, after I enable the productionMode, this isssue is gone!
Thanks for your suggestions.

dinvlad:
Would second on this as related to #353. Currently, we can use the (scroll) event but then what's the best way to detect that `event.offsetY` is really near the end?

One way is to calculate the row position based on this offset, but that breaks down if we use row detail (due to a different height of some rows). I feel like that's not the most direct way though.

Maybe we could use @ViewChild decorator to read out scrollbar's height, but that still seem a bit hacky. All we really need is the total height of all rows to be exposed somewhere in the component. Then, once that height is reached (perhaps within a margin of error), we can just fetch more data and add it to the rows.

Unfortunately, (page) event also doesn't suit it because it gets fired only after we transitioned to a new page.

dinvlad:
Got it working with `(page)` event by pre-fetching the last page:
https://plnkr.co/edit/3UVBTiDGZzRcwZ9dsyIg
That seems to be the most robust solution so far, at least relying on the current API without pixel calculations.

amcdnl:
Awesome! Can you PR A demo for the community?

brian-singer:
I am very interested in this feature, is there any news?

dinvlad:
[...]

PR #740, opened 2017-05-06
Add server-side scrolling demo

dinvlad:
This demo combines scrolling and server-side paging
to enable "infinite" scrolling with results fetched from
the server as the user scrolls down the list.

We always fetch the next page such that it is obvious
from the vertical scrollbar when there are more results.
This approach also helps us avoid explicit pixel calculation logic,
instead relying only on (page) events.

Once loaded, the results are never removed from the cache,
which increases memory consumption over time.
This is a minor hindrance, since in a typical scenario
(e.g. generation of the list in response to a "search" request),
we expect the most relevant results to appear at the top,
with progressively more "stale" results at the bottom,
so that a typical user won't scroll too far (e.g. beyond 100-200 results).
It is also well-known that such search queries consume
progressively more server resources and are not efficient
beyond e.g. 1000 results.

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Demo

There are separate demos for virtual scrolling (when all of the results are fetched in advance)
and server-side paging (without vertical scrolling).

**What is the new behavior?**
We combine these together to enable "infinite" server-side scrolling.

- [ ] Yes
- [x] No

**Other information**:
This PR was requested by @amcdnl in https://github.com/swimlane/ngx-datatable/issues/611#issuecomment-299452014

### DIFF
--- a/demo/app.component.ts
+++ b/demo/app.component.ts
@@ -62,6 +62,7 @@ import { Location, LocationStrategy, HashLocationStrategy } from '@angular/commo
             <ul>
               <li><a href="#client-paging" (click)="state='client-paging'">Client-side</a></li>
               <li><a href="#server-paging" (click)="state='server-paging'">Server-side</a></li>
+              <li><a href="#server-scrolling" (click)="state='server-scrolling'">Scrolling server-side</a></li>
               <li><a href="#virtual-paging" (click)="state='virtual-paging'">Virual server-side</a></li>
             </ul>
           </li>
@@ -128,6 +129,7 @@ import { Location, LocationStrategy, HashLocationStrategy } from '@angular/commo
         <!-- Paging -->
         <client-paging-demo *ngIf="state === 'client-paging'"></client-paging-demo>
         <server-paging-demo *ngIf="state === 'server-paging'"></server-paging-demo>
+        <server-scrolling-demo *ngIf="state === 'server-scrolling'"></server-scrolling-demo>
         <virtual-paging-demo *ngIf="state === 'virtual-paging'"></virtual-paging-demo>
 
         <!-- Sorting -->
--- a/demo/module.ts
+++ b/demo/module.ts
@@ -26,6 +26,7 @@ import { FooterDemoComponent } from './basic/footer.component';
 // -- Paging
 import { ClientPagingComponent } from './paging/paging-client.component';
 import { ServerPagingComponent } from './paging/paging-server.component';
+import { ServerScrollingComponent } from './paging/scrolling-server.component';
 import { VirualPagingComponent } from './paging/paging-virtual.component';
 
 // -- Sorting
@@ -67,6 +68,7 @@ import { ColumnPinningComponent } from './columns/pinning.component';
     RowDetailsComponent,
     ClientPagingComponent,
     ServerPagingComponent,
+    ServerScrollingComponent,
     ClientSortingComponent,
     DefaultSortingComponent,
     ServerSortingComponent,
--- a/demo/paging/scrolling-server.component.ts
+++ b/demo/paging/scrolling-server.component.ts
@@ -1,0 +1,72 @@
+import { Component } from '@angular/core';
+import {MockServerResultsService} from "./mock-server-results-service";
+import {PagedData} from "./model/paged-data";
+import {CorporateEmployee} from "./model/corporate-employee";
+import {Page} from "./model/page";
+
+@Component({
+  selector: 'server-scrolling-demo',
+  providers: [
+      MockServerResultsService
+  ],
+  template: `
+    <div>
+      <h3>
+        Server-side Scrolling
+        <small>
+          <a href="https://github.com/swimlane/ngx-datatable/blob/master/demo/paging/server-scrolling.component.ts" target="_blank">
+            Source
+          </a>
+        </small>
+      </h3>
+      <ngx-datatable
+        class="material"
+        [rows]="rows"
+        [columns]="[{name:'Name'},{name:'Gender'},{name:'Company'}]"
+        [columnMode]="'force'"
+        [headerHeight]="50"
+        [footerHeight]="0"
+        [rowHeight]="50"
+        [scrollbarV]="true"
+        (page)='onPage($event.offset)'>
+      </ngx-datatable>
+    </div>
+  `
+})
+export class ServerScrollingComponent {
+  
+  lastPage = new Page();
+  rows = new Array<CorporateEmployee>();
+
+  constructor(private serverResultsService: MockServerResultsService) {
+    this.lastPage.pageNumber = -1;
+    this.lastPage.size = 5;
+  }
+
+  ngOnInit() {
+    this.addPage(0);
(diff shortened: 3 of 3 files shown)


amcdnl:
Awesome!

How is this different from this example? https://github.com/swimlane/ngx-datatable/blob/master/demo/paging/paging-virtual.component.ts

dinvlad:
I noticed that example (was it added only recently though?), but to the users it appears to load everything in advance (judging by the full 100 count) - even though the code paging service does return pages in pieces, it still assigns `totalElements` which normally we would not have access too. My example queries the server really only for a small chunk in the beginning (so we won't know how many total elements there are in advance), and re-querying for the next block only if needed, when the user scrolls further (so it never uses `totalElements`).

Also, I think there's a problem with that example in
```
rows.splice(start, 0, ...pagedData.data);
```
which adds more pages on every (page), so the list grows way beyond 100 elements over time. I think it should be
```
rows.splice(start, pagedData.data.length, ...pagedData.data);
```
to achieve the effect of re-querying each chunk to avoid "stale" data (which the 'scrolling' example doesn't on purpose).

So in my version we don't have access to all of the potential elements in advance, rather we load and cache as necessary on demand, and don't display "pages" to the user at all. The net effect is the appearance of "infinite" scroll as opposed to paging (which virtual or not appears the same to the user, unless we remove 'count').

amcdnl:
Yes, it was added recently. The fetch and then pluck was only out of demo purposes. Sounds like your demo is more complete so we will use that one instead :)

dinvlad:
Thanks! Let me know if you need anything else.

BBaysinger:
A huge thanks for this. And just in time. I almost didn't find it as the demo site isn't updated with this yet.

dinvlad:
Actually, I'm not completely satisfied with this solution either, since it requires pre-fetching a number of results that is greater than the viewport can host (so that the scrollbar is always there). So we have to set a large static value that will work against all browsers and window sizes.

On smaller screens/windows, we end up pre-fetching far more rows than the client will need, and it only worsens over time. Indeed, a `(page)` event fires once we get past the number of rows that a single viewport can host, which is less than the `limit` set for external paging (it wasn't done in this example, but it makes no difference). I.e. the `pageSize` that triggers `(page)` event is [calculated ](https://github.com/swimlane/ngx-datatable/blob/master/src/components/datatable.component.ts#L905)based on the height of the table / `rowHeight`. So we end up fetching more and more "background" pages over time, and the scrollbar position approaches the ratio of `pageSize` / `limit`.

An alternative to that would be to use the `(scroll)` event, something like
```
@Component({
  selector: 'my-app',
  template: `
    <ngx-datatable
      class='material'
      [rows]='rows'
      [columnMode]="'force'"
      [headerHeight]="headerHeight"
      [rowHeight]="rowHeight"
      [scrollbarV]="true"
      [loadingIndicator]="isLoading"
      (scroll)="onScroll($event.offsetY)"
    >
    ...
  `
})
export class App implements OnInit {
  
  readonly headerHeight = 50;
  readonly rowHeight = 50;
  
  rows: any[] = [];
  isLoading: boolean;
  
  constructor(private el: ElementRef) { }
  
  ngOnInit() {
    this.onScroll(0);
  }
  
  onScroll(offsetY: number) {
    const rowsHeight = this.el.nativeElement.getBoundingClientRect().height - this.headerHeight;
    if (!this.isLoading && offsetY + rowsHeight >= this.rows.length * this.rowHeight) {
      this.loadPage(this.lastPage, Math.ceil(rowsHeight / this.rowHeight));
    }
  }
  
  loadPage(pageIndex: integer, pageSize: integer) {
    this.isLoading = true;
    this.fetch(pageSize).then(results => {
      this.rows.push(...results);
      this.isLoading = false;
    });
  }
}
```
This approach ensures that the next "page" is loaded only once we reach the end of the current view, and the number of results on that "page" is equal to the number that would fit on the current view. Now, if the user resizes the view (either enlarging or shrinking), next time the `(scroll)` event will take into account the new height of the view. Also, if the user scrolled the current view prior to enlarging it, the (scroll) event fires again upon resize and will fetch new results accordingly. The only time it fails is upon initial page load, where if you enlarge it the `(scroll)` event would not fire, unless you try to scroll first. Also, if any row detail is expanded, it will fire earlier in the process but that doesn't hurt as long as the number of expanded rows is small.

Overall this works much nicer and is closest to the experience of 'infinite scrolling' that people are used to. There could be further improvements to handle `(scroll)` events better. But at least, this solution relies on the public API and doesn't use any internals.

amcdnl:
@dinvlad - should we remove the other demo in favor of this one?

dinvlad:
Yes, have just submitted #839

Matmo10:
Ended up here after reading various threads asking about infinite scrolling (with data that comes from a server for each "page"). 

First, I think it would be very helpful to clearly explain somewhere the difference(s) between "Scrolling server-side" and "virtual server-side".  I'm sure these terms are totally obvious to some people, but it's (probably) unclear to some people, myself included. 

Second, the [virtual server side](https://swimlane.github.io/ngx-datatable/#virtual-paging) demo is currently broken (`Error: query at index -1 failed: Fenwick tree array not initialized.`), which makes it even harder to distinguish between the two. 

I would imagine my use case is fairly common: when I approach the bottom of the table, I'd like to get some event that would allow me to go manually fetch the next page and add the rows to my [rows] input.  Looking at the source for the virtual server side pagination example, it seems like thats what I want, but any usage of `[externalPaging]="true"` causes the same error that the demo app is experiencing. 

TL;DR - The demo / code for "virtual paging server side" is broken.

dinvlad:
@Matmo10 sorry, that was a while ago (it worked then) and I'm not sure what the demo page is doing now. I've actually moved away from Angular altogether (to Vue/Vuetify), so not sure I can help any futher.

@amcdnl may probably provide some pointers. My use case is as follows:
1) _virtually_ scroll the table, which means that _only_ the rows in the current browser _view_ are _rendered_, to avoid an ever-growing list of HTML elements that the user sees only a portion of at any given moment;
2) capture an event when the view reaches the end of the table;
3) load more rows from the server and append them to the table; alternatively, load _all_ of 1..N rows (where N is the _new_ "length" of the table) and replace the table with these rows. The latter approach addresses the issue of inconsistent updates between paging requests (so you don't get duplicate/missing rows due to the fact that the data may have been modified on the server between the paging requests). This approach also maps naturally to server-side paging mechanisms in many search engines (if you use one instead of a database), where even if you request rows M..N it will still effectively query all 1..N rows, so there's no benefit of asking only for M..N other than response size.
4) In both cases, ngx-datatable should automatically re-render the virtual view at the current position.

Matmo10:
Thanks for the info. Sounds like my use case is identical - maybe I should try some older versions.

I thought the `virtualization` @Input is what controlled whether the table utilized virtualization or not, though. I'm still not sure why one of the server side pagination examples is considered virtual(ized) while the other is not since neither change the `virtualization` @Input.

Issue #681, opened 2017-04-14
Sorting not working on using 'Template', 'Flex', on using flex for column, resizing is an issue


ichntn:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here
```

**Current behavior**

**Expected behavior**

**Reproduction of the problem**

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

* **Table version:** 7.3.0

* **Angular version:** 4.0.2

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

chintan11192:
I have tried all the possible solutions mentioned in previous issues, hence no luck.
Plus, the performance of the table is also an big issue.
Please suggest some approach or an fix?

amcdnl:
Can you add a demo? or code?

mikevandonselaar:
Make sure you put `let-sort="sortFn"` in your ng-template and make sure to encapsulate your content in a class with a span  `<span (click)="sort()">Your Content here </span>`

This was the case for me. Here's a full code example of my column template:

`<ngx-datatable-column name="Description">
                   <ng-template let-column="column" let-sort="sortFn" ngx-datatable-header-template>
                   <span (click)="sort()">Your Content Here</i></span>Beschrijving
                  </ng-template>
                </ngx-datatable-column>`

ktriek:
I have  a similar  issue, the sort works but no sort (asc/desc) icons and I get the text cursor instead of  the pointer
here is my code
```
<ng-template let-column="column" let-sort="sortFn" ngx-datatable-header-template>
      <span (click)="sort()">{{column.name}}</span>
      <input type="text" (keyup)="onGridFilter($event, column)" class="form-control">
</ng-template>
```

chintan11192:
Thanks,
solved sorting issue by removing header-cell template, i was using multiple column templates.
performance also got enhanced on doing so.

chintan11192:
'flex' for column, resizing the column has no effect on the table only table header gets resized not the rows.

chintan11192:
Still an issue in latest release 9.3.0.

chintan11192:
[...]

PR #742, opened 2017-05-09
Add Datatable SubHeader

ktriek:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Header sorting issue when using custom template #681 

**What is the new behavior?**
Adds a sub header cell, Allows adding a description to column headers, column filtering ... 

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/demo/app.component.ts
+++ b/demo/app.component.ts
@@ -46,6 +46,7 @@ import { Location, LocationStrategy, HashLocationStrategy } from '@angular/commo
               <li><a href="#css" (click)="state='css'">CSS Classes</a></li>
               <li><a href="#dark" (click)="state='dark'">Dark theme</a></li>
               <li><a href="#footer" (click)="state='footer'">Footer Template</a></li>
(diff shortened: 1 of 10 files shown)


amcdnl:
Cool! Question for you: was there any reason why you didn't just use header templates to accomplish this?

ktriek:
@amcdnl  Column sorting doesn't work currently when using header templates,
Also useful when the user wants to keep the default header and just add some details.

amcdnl:
I think a better solution would be to fix the column sorting rather than implement a new header template type.

dzfoxos:
This is very useful in my case, I just want to add column filtering so I don't think I should rewrite the whole header every time.

amcdnl:
@ktriek - Thanks so much for your contribution. I think I'm going to close this PR in light of the approach I suggested. Its more generic and you can achieve the same results.

@dzfoxos - I understand thats a tad inconvenient but I think rather than add extra code, extra edge cases its the right approach for the tool.

thanks again for your contributions!!

Issue #769, opened 2017-05-22
Checkbox selection with shift appears to have some defects


dalelotts:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```
**Current behavior**
1. Go to http://swimlane.github.io/ngx-datatable/#chkbox-selection
1. Check the first row (Ethel)
1. Shift-click on the last row (Georgina)
1. Note that all rows are selected and also listed in the selections list in the upper right
1. Again, Shift-click on the last row (Georgina) 
1. No rows are selected 
1. Defect manifested here --> Georgina is still in the selections list in the upper right.

This is easily reproducible with multiple variations on the shift-click selection/deselection of the check boxes.

**Expected behavior**
Georgina should not be in the selection list OR Georgina's checkbox should be checked.

**Reproduction of the problem**
This is easily reproducible with multiple variations on the shift-click selection/deselection of the check boxes.

This was discovered because I am writing unit tests for selection (PR coming as soon as I am done with the tests) this code is currently completely untested. I would like to understand the observed behavior when combining shiftKey with checkbox selection so I can write logical unit tests for the behavior (I'm ignoring the implementation details as much as possible - single select is done and have not yet started on the other selection types) 

**What is the motivation / use case for changing the behavior?**
I just want the behavior to be logically consistent, whatever it is. At this point, I cannot tease out an explanation or set of unit tests for the currently observed behavior.

**Please tell us about your environment:**
MacOS Seirra 
Node 7.10.0

* **Table version:** 0.8.x
9.1  - I think.Grabbed the latest from master last week. Not exactly working on a fork or the current repo so at the moment I'm not 100% sure of the changes we have from your repo

* **Angular version:** 2.0.x
Angular 4.1

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

My god! Is anyone still using bower?
Chrome 57.0.2987.133 (64-bit)

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
TS 2.2.1 compiled to es2015

amcdnl:
Closing given PR merged

PR #778, opened 2017-05-25
fix(datatable-selection): shift-select behavior is now like finder/explorer

dalelotts:
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

1. #769 
1. command+click on mac does not work
1. shift-click does not work correctly on all selection types

**What is the new behavior?**
1. #769 
1. command+click on mac works (I see this is already fixed)
1. shift-click works correctly on all selection types (no change to single-select)

- [ ] Yes
- [?] No
I don't think so, removed `selectEnabled` since it is not needed.

**Other information**:

Thanks for all of hard work!

-- Your Friends At Veritas 😄


dalelotts:
This PR breaks selection but the tests did not pick it up because the `selected` reference must be passed down from a host component and the footer later modifies that reference. 

Closing since this cannot be accepted as is.

PR #848, opened 2017-07-06
Prevent mutating original rows

mxchange:
- [x] Bugfix
- [x] Feature

Currently `$$index` is set on the original object you pass to `rows`, because it's set by reference, thus polluting the original data.

**What is the new behavior?**
Now it makes a clean copy that it can modify internally

- [ ] Yes
- [x] No

**Other information**:
The original data should never be touched by the table.

### DIFF
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -324,7 +324,7 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
     const temp: any[] = [];
 
     while (rowIndex < last && rowIndex < this.rowCount) {
-      const row = this.rows[rowIndex];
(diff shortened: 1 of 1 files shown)


kyranjamie:
Relates to: https://github.com/swimlane/ngx-datatable/issues/610

amcdnl:
Problem is now if you change a row's value via inline editing for example the original row is not updated.

mxchange:
@amcdnl hmm, then I think the inline editing should be updated to account for updating the cloned row?

amcdnl:
The indexes need to be tracked separately.

amcdnl:
Resolved :)

Issue #891, opened 2017-08-04
Feature request: Turn off <span title=""> rendering in cells


Bidthedog:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[*] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**

I have built a filter mechanism using ngx-datatable that allows users to search individual columns. On top of this, as the user types a search term, I am looping through the rows in targeted column and inserting HTML into the cell values in order to highlight particular pieces of text. Looks awesome, until you mouse-over the cell and you see the inserted span tag:

![ngx-datatable title issues](http://i.imgur.com/tFGVDhW.gif "ngx-datatable example")

It seems that the cell's value is inserted directly into the cell's <span title=""> tag, which is as expected.

The cell content is updated as follows:
```
    for (const row of this.rows) {
      let currentText = row[filterColumn] as string || '';
      currentText = DataTableComponent.stripHtml(currentText);
      row[filterColumn] = this.highlightWords(currentText, searchTerm);
    }
```

**Expected behavior**

**Reproduction of the problem**

**What is the motivation / use case for changing the behavior?**

One of the following solutions would be great:

1. Strip all HTML from the <span title=""> value before its inserted by default
2. Enable / disable the <span title=""> tag in each cell with a column input:
```
this.columns = [
      { prop: 'CustomerId', name: 'Customer Id', enableTitle: false }
];
```
or
```
  @ViewChild(DatatableComponent)
  public table: DatatableComponent;

  myFunct() {
    this.table.columns['CustomerId'].enableTitle = false;
  }
```
3. Provide a way to target rendered cell data (so the title tag can be removed or changed manually _per cell_)

**Please tell us about your environment:**

Windows 8.1, VSCode, angular-cli

* **Table version:** 0.8.x

9.3.1

* **Angular version:** 2.0.x

```
$ ng -v
    _                      _                 ____ _     ___
   / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
  / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
 / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
/_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
               |___/
@angular/cli: 1.2.3
node: 6.9.4
os: win32 x64
```

[...]

PR #892, opened 2017-08-04
#891 Stripped HTML from cell title popup

Bidthedog:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

See #891

**What is the new behavior?**

HTML tags are stripped from the `<span title="">` value inserted into each cell.

- [ ] Yes
- [ ] No
- [x] Maybe

It's possible that someone would want the HTML to be displayed in the cell's title popup, though I find it hard to think of a use-case for that?

**Other information**:


Bidthedog:
Second commit is not necessary, was just trying to help with the build!

Issue #785, opened 2017-05-30
headerTemplate creates new outletContext on every change


arlowhite:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**

I use `headerTemplate` to put a material2 `MdInputContainer` in my column headers to provide a filter UI.
The `headerTemplate` regenerates far too often (on every change detection).  This resets the input text in my template as the user types. I worked around this by saving the input text externally, but even then there are serious issues.
In Firefox, when I select all text in the input, the entire Firefox browser crashes in Linux, and the tab crashes in Firefox Mac. I think elements are being destroyed while Firefox has a pointer to them in its text selection code.

Currently, the `headerTemplate` outlet context is created in the template like this:
```typescript
      <ng-template
        *ngIf="column.headerTemplate"
        [ngTemplateOutlet]="column.headerTemplate"
        [ngOutletContext]="{ 
          column: column, 
          sortDir: sortDir,
          sortFn: sortFn,
          allRowsSelected: allRowsSelected,
          selectFn: selectFn
        }">
      </ng-template>
```

I think this creates a new `ngOutletContext` object on every change detection, which causes the `headerTemplate` to be recreated.

**Expected behavior**

`headerTemplate` should only regenerate when properties it depends on change.

**Reproduction of the problem**

I don't have time to create a stand-alone demo at the moment.

**What is the motivation / use case for changing the behavior?**

Prevent `headerTemplate` from regenerating on every change detection.

**Please tell us about your environment:**

Node v6.9.1
npm 4.6.1

* **Table version:** 9.2.0

* **Angular version:** 4.1.3

```
 @angular/cli: 1.0.4
 node: 6.9.1
 os: linux x64
 @angular/animations: 4.1.3
 @angular/common: 4.1.3
 @angular/compiler: 4.1.3
 @angular/core: 4.1.3
 @angular/flex-layout: 2.0.0-beta.8
 @angular/forms: 4.1.3
 @angular/http: 4.1.3
 @angular/material: 2.0.0-beta.5-22a55aa
 @angular/platform-browser: 4.1.3
 @angular/platform-browser-dynamic: 4.1.3
 @angular/router: 4.1.3
 @angular/cli: 1.0.4
 @angular/compiler-cli: 4.1.3
```

* **Browser:** all

* **Language:** all

[...]

PR #903, opened 2017-08-08
prevent headerTemplate outlet context from regenerating on every change.

arlowhite:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

#785 

**What is the new behavior?**
Modified **DataTableHeaderCellComponent**
A new outlet context object will only be created when one of its properties changes. Triggered by ngOnChanges.
Also added `changeDetection: ChangeDetectionStrategy.OnPush`

- [ ] Yes
- [x] No

If this PR contains a breaking change, please describe the impact and migration path for existing applications: 

Shouldn't break existing code unless the OnPush change detection change causes issues for people who are mutating objects instead of setting Inputs.

**Other information**:

### DIFF
--- a/src/components/header/header-cell.component.ts
+++ b/src/components/header/header-cell.component.ts
@@ -1,9 +1,26 @@
 import {
-  Component, Input, EventEmitter, Output, HostBinding, HostListener
+  Component, Input, EventEmitter, Output, HostBinding, HostListener, ChangeDetectionStrategy, OnChanges, SimpleChanges,
+  OnInit
 } from '@angular/core';
 import { SortDirection, SortType, SelectionType, TableColumn } from '../../types';
 import { nextSortDir } from '../../utils';
 import { mouseEvent } from '../../events';
+
+/**
+ * properties provided to the headerTemplate
+ */
+interface HeaderTemplateContext {
+  column: TableColumn;
+  sortDir: SortDirection;
+  sortFn: () => void;
+  allRowsSelected: boolean;
+  selectFn: (o: any) => void;
+}
+
+/**
+ * Input properties that if changed, should re-generate the headerTemplateContext
+ */
+const headerContextProps = ['column', 'sorts', 'allRowsSelected'];
 
 @Component({
   selector: 'datatable-header-cell',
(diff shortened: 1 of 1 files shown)


src/components/header/header-cell.component.ts
```diff
@@ -141,11 +153,31 @@ export class DataTableHeaderCellComponent {
...
+  ngOnInit() {
+    this.updateHeaderTemplateContext();
+    this.initialized = true;
+  }
+
+  ngOnChanges(changes: SimpleChanges): void {
+    if (!this.initialized) return;
+    // determine if a property that affects headerTemplateContext has changed
+    for (const prop in headerContextProps) {
+      if (changes.hasOwnProperty(prop)) {
```
amcdnl: Why not just updated the individual changed item vs the whole object?

src/components/header/header-cell.component.ts
```diff
@@ -141,11 +153,31 @@ export class DataTableHeaderCellComponent {
...
+  private initialized: boolean;
+
   @HostListener('contextmenu', ['$event'])
   onContextmenu($event: MouseEvent): void {
     this.columnContextmenu.emit({ event: $event, column: this.column });
   }
 
+  ngOnInit() {
+    this.updateHeaderTemplateContext();
+    this.initialized = true;
```
amcdnl: What is `initialized` used for?

src/components/header/header-cell.component.ts
```diff
@@ -141,11 +153,31 @@ export class DataTableHeaderCellComponent {
...
+  }
+
+  ngOnChanges(changes: SimpleChanges): void {
+    if (!this.initialized) return;
+    // determine if a property that affects headerTemplateContext has changed
+    for (const prop in headerContextProps) {
+      if (changes.hasOwnProperty(prop)) {
+        this.updateHeaderTemplateContext();
```
amcdnl: Why not just updated the individual changed item vs the whole object?

src/components/header/header-cell.component.ts
arlowhite: I think when I originally worked on this, ngOnChanges was called more than once during the Component startup lifecycle.
However this should be reviewed.

src/components/header/header-cell.component.ts
arlowhite: I thought you needed to create a new object to trigger change detection in the headerTemplate.

amcdnl:
@arlowhite - i updated it to the implementation i did with body cell in 10.0.4. Do you wanna check it out and see if it addresses the issues?

arlowhite:
Yes, 10.0.5 fixes this issue. :+1: 

Tested in my project with custom headerTemplate, and template doesn't reset on change detection anymore.

Issue #909, opened 2017-08-14
Support for mouseenter event for a row


szpetny:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[x ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
Mouseenter event handler available on the entire table

**Expected behavior**
Mouseenter event handler available in body-row.component.
Something like:

  @HostListener('mouseenter', ['$event'])
  onMouseenter(event: Event) {
    this.activate.emit({
        type: 'mouseenter',
        event,
        row: this.row,
        rowElement: this.element
      });
  }

**Reproduction of the problem**
Adding (mouseenter)="method()" to ngx-datatable lets handle mouseenter for the entire table, I need to for example log sth every time a row is hovered with mouse. 

**What is the motivation / use case for changing the behavior?**
It would be useful to be able to customize behavior when hovering mouse over the table records. It is a quite common use case. 

**Please tell us about your environment:**
Windows7, node v7.2.1

* **Table version:**
10.0.5

* **Angular version:** 
4.3.0
* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

amcdnl:
Sure. Can you PR this?

szpetny:
yes

szpetny

14 sie 2017 16:18 "Austin" <notifications@github.com> napisał(a):

amcdnl:
Merged.

PR #912, opened 2017-08-15
Support for mouseenter event for a row

szpetny:
- [ ] Bugfix
- [x ] Feature
- [ ] Build related changes
- [ ] CI related changes

https://github.com/swimlane/ngx-datatable/issues/909

**What is the new behavior?**
Mouseenter event handler available in body-row.component.

- [ ] Yes
- [ x] No

**Other information**:


amcdnl:
Hmmm, activate event. I think of activate when it has focus via click or keyboard. this almost needs a new event type.

szpetny:
dunno, for me hovering is also kind of activation ಠ_ಠ but anyways I need this feature, so want me to introduce a new emitter?

amcdnl:
Ya, make sense. No I'll go with this.

bmcleod:
Thanks for adding this feature! Is it possible to get an example on how to use it/documentation?

szpetny:
You need to bind an event handler `<ngx-datatable (activate)="onActivate($event)">
        </ngx-datatable>` then inside this event handler do a check `onActivate(event) {
if (event.type === 'mouseenter') {
 console.log(event);
}
  }`,
 cheers

wizarrc:
Just a word of caution, this may break people who don't check the event type but depend somehow on the event to maintain some kind of state. An earlier project I worked on did just that.

@amcdnl might be a good idea to add to the changelog that you check the event type if you need to discriminate against hover activation.

wizarrc:
@szpetny Sorry for being late on the thread. I like the idea of having many forms of activation, including mouse hovering. I can anticipate cool new ways of interacting with the table. Interesting idea!

I'd be curious from a performance point of view. Imagine for a moment that you are scrolling via the scroll wheel on the mouse. Since the activate event is being listened to on the angular zone, it will trigger a change detection upon handling the event. Now the question is will that event trigger many times during a scroll, causing performance to suffer? If so, it might be a good idea to register the event outside of angular zone, filter out any events during scrolling operations, then if there are any listeners, handle the event back in the angular zone, with full change detection logic.

Prior to this, css would be the only way to style for hovering not causing any change detection problems.

szpetny:
perhaps there should be some delay to avoid too frequent firing, probably could look at it at the weekend

szpetny:
Ok, I committed sth, not sure if you like it bc it introduces another listener (to mouseleave). 
Anyways this is the fork:
https://github.com/szpetny/ngx-datatable

wizarrc:
@szpetny the problem with counting on the `mouseleave` event is that it doesn't always trigger. An example of this is during debugging, and setting a break point. Another example where it doesn't trigger the event is when the OS displays another window for whatever reason, and the browser looses focus, then when the browser gets focus again, the mouse has already moved thus preventing the `mouseleave` event from triggering.

wizarrc:
@szpetny related issue https://github.com/swimlane/ngx-datatable/issues/989

I think the perf is fine for now though. If you want to make any perf changes, I think something like this would be preferred.

datatable-body.component.ts
```
ngOnInit() {
    this.zone.runOutsideAngular(() => {
      this.element.addEventListener('mouseenter', this.onMouseenter.bind(this));
    });
}
ngOnDestroy() {
  this.element.removeEventListener(this.onMouseEnter.bind(this));
}
```
datatable.component.ts
```
<datatable-body
...
(activate)="onActivate($event)"
...
onActivate(event) {
  if(this.activate.observers.length) {
    this.zone.run(() => this.activate.emit(event));
}
```

wizarrc:
Unfortunately the mere listening to the event on the datatable component `(activate)="onActivate($event)"` will mark change detection to check but will not actually do the detection till next change detection, which may or may not fire, causing a weird inconsistency that could cause hard to find bugs.

An even better approach is to somehow know if there are any listeners in the main datatable component from the body component and only then trigger the event in the body component. That would mean there would have to be yet another `@input()` presenting that data to the body component. That means yet another change detection variable to check during change detection.

FYI, I solved this problem in my fork by gutting out all this state transfers between components and using a service provider between components, but that's a whole other topic. Needless to say, it gets complicated to make perf improvements without restructuring the internals.

szpetny:
Aww, didn't know that mouseleave is unreliable. Then I guess let it be as it is, apparently the straightforward solutions are the best sometimes..

wizarrc:
@szpetny I updated my recommendation in comment https://github.com/swimlane/ngx-datatable/pull/912#issuecomment-331716695

wizarrc:
@szpetny yeah, straightforward can be less buggy, but using the built-in angular constructs can be a perf problem, but to solve the issue, it can get quite complicated, and hard to find all edge cases, as you can see in my spitball example.

szpetny:
Yes, so making a conclusion, you said it might be a performance problem, then you did some tests and said it is actually ok, though it could be more neat, then I checked my new idea and it failed, then you came out with an idea which is sophisticated, but also complicated and can cause hard-to-find bugs. So I understand: let's leave it as it is? At least for a while

wizarrc:
@szpetny I'm pretty sure my latest refresh on my code example solves the hard to find bug and event consistency issue, but it's still hard to follow for those not sure how the angular eventing system, zones, and change detection works internally. I vote to leave it alone till the internals are better structured.

Issue #943, opened 2017-08-30
All check box and Radio buttons in app are hidden due to css


anthonysubramani:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**

All check Bo
**Expected behavior**

All check box and Radio buttons in app are hidden due to this css 
`input[type="checkbox"],
input[type="radio"] {
  display: none; }
` in file
@swimlane/ngx-datatable/release/themes/material.css
**Reproduction of the problem**

checkbox not display in demo
http://swimlane.github.io/ngx-datatable/#chkbox-selection
**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

* **Table version:** 0.8.x

* **Angular version:** 2.0.x

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

Debananda:
Adding to the problem : As the checkbox styling is given on a generic way (without defined inside a class) it also affects the whole consuming project

tnarp:
`
input[type="checkbox"], input[type="radio"] {
    display: none;
}
`
modify to 
`
input[type="checkbox"], input[type="radio"] {
    display: none;
}
`
then the checkbox can shwo

anthonysubramani:
@tnarp the solution is evident, but this needs to be changed in the source itself. Also inputs are inline element so adding `display: block !important` will make it even worse. so it should be associated with `inline` style

monica7l:
I agree this has to be fixed in ngx-datatable itself. In the mean time I closed my #952 because it was a duplicate of this one.

Issue #939, opened 2017-08-29
checkbox not show


hevan:
**I'm submitting a ...**  (check one with "x")
```
[ x] bug report => search github for a similar issue or PR before submitting
[ x] feature request
[ x] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**

**Expected behavior**

**Reproduction of the problem**

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

* **Table version:** 0.8.x

Angular version:** 2.0.x
checkbox not show

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

* **Language:** [all | TypeScript 2.4.2 |  ES5]

Jackson-soft:
same issue 

ngx-datatable : 10.2.0
TypeScript : 2.4.2

hevan:
this problem  Affects all of the checkbox components。
all page , all checkbox not show!!

ngx-datatable : 10.2.0
TypeScript : 2.4.2

anthonysubramani:
Delete the followng css present at top of the file @swimlane/ngx-datatable/release/themes/material.css in node modules
`input[type="checkbox"], input[type="radio"] { display: none; }`

hevan:
Thank you!

mtraut:
Many thanks, was a showstopper

PR #956, opened 2017-09-05
fix checkbox display

Jotakuun:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

#943 

**What is the new behavior?**
I just removed the display none from the material.scss.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -682,7 +682,8 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
    * Gets the row index given a row
    */
   getRowIndex(row: any): number {
-    return this.rowIndexes.get(row) || 0;
+    // return this.rowIndexes.get(row) || 0;
+    return this._rows.indexOf(row);
   }
 
 }
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -682,8 +682,7 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
    * Gets the row index given a row
    */
   getRowIndex(row: any): number {
-    // return this.rowIndexes.get(row) || 0;
-    return this._rows.indexOf(row);
+    return this.rowIndexes.get(row) || 0;
   }
 
 }
--- a/assets/app.css
+++ b/assets/app.css
@@ -242,6 +242,11 @@ select {
   color: #72809b;
 }
 
+.expectedpayment, .expectedpayment2, .expectedpayment3
+{
+  display: none;
+} 
+
 .expectedpayment + label:before {
   border: 1px solid #7f83a2;
   content: "\00a0";
--- a/src/themes/material.scss
+++ b/src/themes/material.scss
@@ -1,6 +1,3 @@
-//#c1,#c2,#c3
-//.expectedpayment_checkbox
-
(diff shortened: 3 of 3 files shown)


Jotakuun:
I don't know why checkbox and radio inputs wasn't displaying because if there is a reason to hide it or a new behaviour just tell me and I'll fix it and update the pull request.

wizarrc:
@Jotakuun Line 1 - 8 should never have been added. It looks like it was accidently added from the grouping rows PR. That stuff should be moved to the demo css file.

wizarrc:
I suggest moving line 1-8 to https://github.com/swimlane/ngx-datatable/blob/master/assets/app.css next to `.expectedpayments` section, and uncomment line 1 and 2.

Jotakuun:
Ok, thank you @wizarrc 
I removed the comments I left in material.scss and I moved those lines to the app.css 

By the way, those classes and ids selectors doesn't match to the row-grouping checkboxes anymore. So I named it like they are now. But anyway, row-grouping demo is not working because all labels are recieving same name and id from rowIndex. I'm going to investigate what's happening and let's see if I can fix it.

src/components/body/body.component.ts
```diff
@@ -682,7 +682,8 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
    * Gets the row index given a row
    */
   getRowIndex(row: any): number {
-    return this.rowIndexes.get(row) || 0;
+    // return this.rowIndexes.get(row) || 0;
```
wizarrc: @Jotakuun The `indexOf` array function is much slower than the Map `get` function. Are you having issues with it returning the wrong value?

Jotakuun: Ok, so it looks like the getRowIndex() was changed in last release and this piece of code wasn't working for rowGroups because rowIndexes is not returning any index for rowGroups. I couldn't figure it out why. Any suggestion?

wizarrc: hmm, not sure. Didn't really get a chance to look over the row groups PR that much. I can tell you that `rowIndexes` should have a list of indexes per row, same as the `rows` array being searched manually, but much slower. If `rowIndexes` isn't producing correct results, then it's not being updated correctly.

wizarrc:
related issue #939

src/components/body/body.component.ts
amcdnl: Grouped row indexes are in a different collection.

Jotakuun: And where is it? The ```rowIndexes``` are not updating in the ```updateRow()``` method when there's groupedRows.

wizarrc: Can we treat this is a separate issue? I find value in moving the css out of the `material.scss` as it's causing issues.

amcdnl: i removed it and released thisx morning.

Jotakuun: I agree, this is completely another issue. I update the PR again with just the ```material.scss``` and we treat this in another pull request.

iamchiranjeevin:
im using latest version i still cant see the checkboxes

PR #968, opened 2017-09-08
Fixes IE & Edge displaying null for no column name

wizarrc:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

IE & Edge display null in header column when column name and property are undefined. Also case where prop is set and name is `0`. Name will be converted to prop.

**What is the new behavior?**

Header column will no longer display null when column name and property are undefined. Also fixes case where prop is set and name is `0`. Name will no longer be converted to prop.

- [ ] Yes
- [x] No

**Other information**:

null

### DIFF
--- a/src/utils/column-helper.ts
+++ b/src/utils/column-helper.ts
@@ -26,8 +26,12 @@ export function setColumnDefaults(columns: TableColumn[]) {
     }
 
     // format props if no name passed
-    if(column.prop != null && !column.name) {
+    if(column.prop != null && column.name == null) {
       column.name = deCamelCase(String(column.prop));
+    }
+    
+    if(column.prop == null && column.name == null) {
+      column.name = ''; // Fixes IE and Edge displaying `null`
     }
 
     if(!column.hasOwnProperty('resizeable')) {
--- a/src/utils/column-helper.ts
+++ b/src/utils/column-helper.ts
@@ -17,7 +17,7 @@ export function setColumnDefaults(columns: TableColumn[]) {
 
     // prop can be numeric; zero is valid not a missing prop
     // translate name => prop
-    if(column.prop == null && column.name) {
+    if(isNullOrUndefined(column.prop) && column.name) {
       column.prop = camelCase(column.name);
     }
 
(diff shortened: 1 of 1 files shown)


src/utils/column-helper.ts
```diff
@@ -26,9 +26,13 @@ export function setColumnDefaults(columns: TableColumn[]) {
     }
 
     // format props if no name passed
-    if(column.prop != null && !column.name) {
+    if(column.prop != null && column.name == null) {
```
amcdnl: Can you convert these to `!==` and `===`

src/utils/column-helper.ts
```diff
@@ -26,9 +26,13 @@ export function setColumnDefaults(columns: TableColumn[]) {
...
       column.name = deCamelCase(String(column.prop));
     }
+    
+    if(column.prop == null && column.name == null) {
```
amcdnl: Same as above w/ conversion.

src/utils/column-helper.ts
wizarrc: I can but it will take up twice as much code to check for null and undefined. Is that what you want?

src/utils/column-helper.ts
```diff
@@ -52,6 +56,10 @@ export function setColumnDefaults(columns: TableColumn[]) {
   }
 }
 
+export function isNullOrUndefined<T>(value: T | null | undefined): value is null | undefined {
```
wizarrc: Making a function here seems more readable. In the future, this can be moved to a more general purpose area.

wizarrc:
I made a helper function to handle checking for both null and undefined. That should pass tslint.

src/utils/column-helper.ts
amcdnl: We do this strict checking everywhere else, i'd like to keep it consistent unless that type of checking is needed.

wizarrc: Makes sense. I saw it wasn't using it here so I kept it that way with my changes. In fact, my last PR didn't do strict checking as well, for the other IE/Edge issue. I will do strict checking for changes going forward.

PR #975, opened 2017-09-11
Doesn't delete row expansions when changing rows

wizarrc:
I tried not to touch row grouping other than checking the size (WeakMap doesn't have capability), but that may still be buggy.

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Every time you refresh rows i.e. `this.rows = [...this.rows];` like you suggest, it clears the list of expanded rows. But if you don't clear the list of expanded rows, then you could have a memory leak. I would recommend using a WeakMap.

**What is the new behavior?**

Expanded Rows mapping is not deleted until no longer referenced.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -119,7 +119,7 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
 
   @Input() set rows(val: any[]) {
     this._rows = val;
-    this.rowExpansions.clear();
+    this.groupsInit = false;
     this.recalcLayout();
   }
 
@@ -216,7 +216,8 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
   rowTrackingFn: any;
   listener: any;
   rowIndexes: any = new Map();
-  rowExpansions: any = new Map();
+  rowExpansions: any = new WeakMap();
+  groupsInit: boolean = false;
 
   _rows: any[];
   _bodyHeight: any;
@@ -601,7 +602,8 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
    */
   toggleAllRows(expanded: boolean): void {
     // clear prev expansions
-    this.rowExpansions.clear();
+    this.groupsInit = false;
+    this.rowExpansions = new WeakMap();
 
     const rowExpanded = expanded ? 1 : 0;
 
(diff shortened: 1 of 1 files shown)


amcdnl:
If you do `[...this.rows]`, the row object is still the same object, the array has just changed so the map should stay valid.

wizarrc:
@amcdnl I'm pretty sure what will happen is when you do `this.rows =` anything such as `this.rows = [...this.rows];` you will trigger the datatable rows setter

https://github.com/swimlane/ngx-datatable/blob/master/src/components/datatable.component.ts#L111
```
@Input() set rows(val: any) {
...
    if (!this.externalSorting) {
      this._internalRows = sortRows(val, this._internalColumns, this.sorts);
    } else {
      this._internalRows = [...val];
    }
...
```
The datatable rows setter assigns a new `_internalRows` value (not modifies existing one). The body component is binded to the `_internalRows` property
```
<datatable-body
...
    [rows]="_internalRows"
...
```

which will trigger the body rows setter

https://github.com/swimlane/ngx-datatable/blob/master/src/components/body/body.component.ts#L120
```
...
    this.rowExpansions.clear();
...
}
```

The `rowExpansions` will be cleared every update!

wizarrc:
@amcdnl Here is an example where setting an array with the exact same items will still trigger the component's setter property. Change detection only does a shallow check, and in this case, the array object is different on the outside (shallow) even though the inside items are all the same.

https://plnkr.co/edit/CcfaybnRYKTl4yCmZjiv

amcdnl:
@wizarrc - Thats exactly what i was saying lol. Since the `rowExpansions` is pointers to the row and the row inside the rows are not actually reset the map shouldn't leak unless u remove / add rows.

wizarrc:
@amcdnl I don't think we are on the same page here. Previously adding or removing rows with `splice` was allowed, whereas now it will not work and you must supply a new array to refresh. When refreshing the rows by adding or removing, or just giving a new array, will flush the `rowExpansions`. Is that what we want to happen after every mutation to the collection?

wizarrc:
Here is a me expanding 2nd and 3rd row and removing first row. You can see the expanded rows are now reset.
remove row

Here is a me expanding 2nd and 3rd row and refreshing the rows (just reassigning them to a new array i.e. `this.rows = [...this.rows];`). You can see the expanded rows are now reset.
refresh rows

http://plnkr.co/edit/YiKBIP1GjpbNOB1K9S28?p=preview

wizarrc:
I'm closing this due to conversation about it being too hacky a fix.

PR #981, opened 2017-09-12
No need to send page event when vertical scrolling is false

vixriihi:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

If table body has scroll and either scrollbarV or scrollbarH it'll trigger page event on scroll.

**What is the new behavior?**
when scrolling body, page event is only emited if scrollbarV is true

- [ ] Yes
- [x] No

**Other information**:
This allows me to have scrollbarH true and scrollbarV false and yet to have body size set with css so that it has a vertical scroll and server side pager. Goal is to have sticky header, horzontal & vertical scroll and server-side pages.

### DIFF
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -335,7 +335,7 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
       offset = Math.ceil(offset);
     }
 
-    if (direction !== undefined && !isNaN(offset)) {
+    if (this.scrollbarV && direction !== undefined && !isNaN(offset)) {
       this.page.emit({ offset });
     }
   }


src/components/body/body.component.ts
```diff
@@ -335,7 +335,7 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
       offset = Math.ceil(offset);
     }
 
-    if (direction !== undefined && !isNaN(offset)) {
+    if (this.scrollbarV && direction !== undefined && !isNaN(offset)) {
       this.page.emit({ offset });
```
wizarrc: @vixriihi This is a page event, not scroll event.

vixriihi: true and it's triggered from onBodyScroll.

wizarrc: I did notice that the scroller component does not turn off scroll events when the `scrollbarV` or `scrollbarH` options changes

```
  ngOnInit(): void {
    // manual bind so we don't always listen
    if (this.scrollbarV || this.scrollbarH) {
      this.parentElement = this.element.parentElement.parentElement;
      this.onScrollListener = this.renderer.listen(
        this.parentElement, 'scroll', this.onScrolled.bind(this));
    }
  }
```

This can be moved and check based on a virtual property, and listen and stop listening based on the property

vixriihi:
This doesn't work as hoped (cannot have many rows since it's really slow). Closing this and adding new PR for better solution

PR #983, opened 2017-09-13
(feature) Add an option to only use pager when using scrollbarv and external pager

vixriihi:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Page change event is triggered from the footer and when scrolling content

**What is the new behavior?**
You can use onlyPagerToChangePage to force paging events to happen only from the footer.

- [ ] Yes
- [x] No

**Other information**:
This allows me to have scrollbarV and externalPager true.
Goal is to have sticky header, horzontal & vertical scroll and server-side pages.

### DIFF
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -335,7 +335,7 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
       offset = Math.ceil(offset);
     }
 
-    if (direction !== undefined && !isNaN(offset)) {
+    if (this.scrollbarV && direction !== undefined && !isNaN(offset)) {
       this.page.emit({ offset });
     }
   }
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -335,7 +335,7 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
       offset = Math.ceil(offset);
     }
 
-    if (this.scrollbarV && direction !== undefined && !isNaN(offset)) {
+    if (direction !== undefined && !isNaN(offset)) {
       this.page.emit({ offset });
     }
   }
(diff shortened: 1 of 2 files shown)


wizarrc:
@vixriihi This doesn't smell right. I'm not sure what is being achieved here.

Why not just turn off `scrollbarV`? How does not having this prevent a stick header?

vixriihi:
Yes I wish I could find a good solution for this but... 
If I turn of scrollbarV page would have a lot of rows at same time -> very slow page

This PR allows me to have body that uses scrollbarV to only render the visible rows and to have the pager in the footer for external server-side paging

wizarrc:
http://plnkr.co/edit/fub1CQpqxMxEuB7exDhU?p=preview has external paging turned on

You could then prefetch the nearby pages. If they jump to it, it starts out displaying blank till you get the page event, fetch the data, modify the array. I have a fork that uses a virtual scrollbar, and that allows all that to happen before it actually performs the scroll to prevent displaying whitespace.

vixriihi:
Here is a plunker to hopefully better describe the problem http://plnkr.co/edit/XiNrMadcEzHISptcJZQt
Pages load nice and fast until you click the last page link in the footer.

wizarrc:
@vixriihi have you tried turning off `scrollbarV` and turning on `limit`? Just get rid of the scroll bar altogether.

vixriihi:
Yes I have. Our customer want's 5000 rows per page and without scrollbarV browsers just cannot handle it.

wizarrc:
ah, yeah, that's a problem. I would suggest you fork it until this can be generalized. That's something I'm working on in my fork but I can't say when I might get that into the library. It was never really designed for that workflow, and only ever added external paging to try to bridge the gap.

vixriihi:
SGTM This PR solves this problem by letting limit to define how many rows will be in the body and then by handling page change events only from the footer. ScrollbarV can easily handle 5000 rows and since there is full set of new rows when page is changed there is no problem of huge master row set with all the rows in it. Here is a plunker to show how it works http://plnkr.co/edit/mLYIvoGHRkGxoOD4iJjU?p=preview 

But I'll close this since more generalized version is on it's way.

Issue #193, opened 2016-10-10
Container resize column recalculation


deeg:
**I'm submitting a ...**  (check one with "x")

```
[ ] bug report => search github for a similar issue or PR before submitting
[ x ] feature request
[ ] support request => Please do not submit support request here
```

**Current behavior**

It seems from [this line](https://github.com/swimlane/angular2-data-table/blob/90ad41d56ec58522b7793aaa6d1f9265ff7ed024/src/components/datatable.component.ts#L198) that the recalculation of column widths only happen on window resize event.

I have a container for a table which gets smaller if you open a filter section next to it. This does not fire that event and the recalculation does not happen.

**Expected behavior**

I would ask that we either bind to a resize event of the data table itself if that is possible, or give access to the [adjustSizes](https://github.com/swimlane/angular2-data-table/blob/90ad41d56ec58522b7793aaa6d1f9265ff7ed024/src/components/datatable.component.ts#L165) function so we can call it programmatically to re calculate the column widths.

**Reproduction of the problem**
http://plnkr.co/edit/wLIdtlUA0ZPrJL7nnpec?p=preview

Notice when clicking smaller/larger buttons to resize the container the columns don't recalculate. If you change the size of the window it will recalculate.

**What is the motivation / use case for changing the behavior?**
Being able to programmatically change the container size of the table, and have the table still render correctly.
- **Table version:** 0.7.x
  0.9.3
- **Angular version:** 2.0.x
  2.0.2
- **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
  Checked Chrome and FF both had this issue.
- **Language:** [all | TypeScript X.X | ES6/7 | ES5]
  TS 2.0.2

I am happy to work on a PR for this if it is something you are open to, but wanted to get feedback first.

amcdnl:
Probably need something like this - http://www.backalleycoder.com/2013/03/18/cross-browser-event-based-element-resize-detection/

I would rather not expose the internal method.

deeg:
Okay thanks, will look into that solution.

ghetolay:
Angular is already doing a throttle so why not just do something like that : 

```
ngOnCheck() {
   if( this.element.offsetWidth != this.width || this.element.offsetHeight != this.height ) {
     this.width = this.element.offsetWidth;
[...]

PR #1033, opened 2017-10-05
Fix visiblility stops check after positive result

wizarrc:
The visibility directive stops check after the first positive result. This is be problematic when visibility toggles between visible and hidden more than once.

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Toggle the table hidden and visible more than once. The body renders incorrectly.

**What is the new behavior?**

Toggle the table hidden and visible more than once. The body renders correctly.

- [ ] Yes
- [x] No

**Other information**:

hidden-resize

### DIFF
--- a/src/directives/visibility.directive.ts
+++ b/src/directives/visibility.directive.ts
@@ -1,5 +1,5 @@
 import {
-  Directive, Output, EventEmitter, ElementRef, HostBinding, NgZone, OnInit, OnDestroy
+  Directive, Output, EventEmitter, ElementRef, HostBinding, NgZone, OnInit, OnDestroy, DoCheck
 } from '@angular/core';
 
 /**
@@ -14,7 +14,7 @@ import {
  *
  */
 @Directive({ selector: '[visibilityObserver]' })
-export class VisibilityDirective implements OnInit, OnDestroy {
+export class VisibilityDirective implements OnInit, OnDestroy, DoCheck {
 
   @HostBinding('class.visible') 
   isVisible: boolean = false;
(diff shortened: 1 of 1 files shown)


wizarrc:
I forgot to mention, steps to reproduce the bug is:

1. Make the table visible.
2. Make the table hidden.
3. Resize the window.
4. Make the table visible again.

amcdnl:
I'm not sure this is the right change. We want to stop checking ASAP since its very pricey and a super hack IMO.

I think a better mechanism would be check on resize if the element is visible and if not just don't resize. You could also start the visibility check at that point again.

wizarrc:
@amcdnl I wasn't sure if it was just an oversight. I noticed the timer myself and since then, I was looking at this other open source project that can detect any div resizing very efficiently and not based on timers.

https://github.com/marcj/css-element-queries

Looking at the internals, it creates a hidden div inside the monitored element, and somehow sets it's width and height very large to where it has a scrollbar for both dimensions. I think the magic happens when it registers a scroll event handler, and I'm guessing that triggers a scroll event on every resize. That might be an interesting replacement. Take a look.

wizarrc:
@amcdnl I updated the PR. I tried to keep the change as minimal as possible.

FYI, I did learn something while looking at the project mentioned above. The visibility directive is primarily used to refresh the table, similar to a windows resize. I thought the above project could essentially replace the visibility directive altogether. That's actually false. The `css-element-queries` project only works when the element is visible. But by having an event that can trigger when it's visible and disable once it's hidden again, this actually would fix the #193 without a user-required workaround or incurring a perf hit. I tried it out, and it looks like it can replace the `@HostListener('window:resize')` event as well. Since the window resize event can unnecessarily trigger refreshes even when the table isn't affected, it actually is more efficient.

Demo of issue 193 that `css-element-queries` can fix efficiently:
sidebar

wizarrc:
I will close this in favor of newer PR that makes this obsolete.

PR #1034, opened 2017-10-05
Fixes #1032

### LINKED ISSUE #1032: Column with name "value" not working
```
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
If column named "value" then render fail with error:
```ERROR Error: Cannot find a differ supporting object 'data1_1' of type 'string'. NgFor only supports binding to Iterables such as Arrays.```

**Expected behavior**
Should work as expected.

```
<ngx-datatable class="material "
            [rows]="csvRows" 
            [columns]="csvColumns"
            [columnMode]="'force'" 
            [headerHeight]="20" 
            [footerHeight]="20" 
            [rowHeight]="20" 
            [loadingIndicator]="loadingIndicator" 
            [limit]="10">
        </ngx-datatable>
```
And somewhere in component.ts:
```
    this.csvColumns.push({name: 'value'},{name: 'test2'});
    this.csvRows.push({'value': 'data1_1', 'test2': 'data1_2'});
    this.csvRows.push({'value': 'data2_1', 'test2': 'data2_2'});
```

This is bug and it's would be great if it will be fixed.

MacOS 10.12.6

* **Table version:** 0.8.x
This worked under 7.x version but appears after update to latest (10.2.3) version

* **Angular version:** 2.0.x
4.3.6

* **Browser:** Safari 11.0

* **Language:** TypeScript 2.4.2

wizarrc:
`groupedRows` should be checked instead of `group.value`. In addition, if `group.value` contains a value from a non-grouped rows, it will try to iterate it. This should fix that by making this an if-else template instead of an if template with an `ngForOf` template.

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

non-grouped rows cannot have `value` as a property.

**What is the new behavior?**

non-grouped rows can have `value` as a property.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -44,7 +44,7 @@ import { mouseEvent } from '../../events';
           [rowIndex]="getRowIndex(group[i])"
           (rowContextmenu)="rowContextmenu.emit($event)">
           <datatable-body-row 
-            *ngIf="!group.value"        
+            *ngIf="!groupedRows; else groupedRowsTemplate"        
             tabindex="-1"
             [isSelected]="selector.getRowSelected(group)"
             [innerWidth]="innerWidth"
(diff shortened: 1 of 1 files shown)


NauxLiu:
We've been bitten by it, can you cut a new release? @achimha

PR #1040, opened 2017-10-07
Resize observer component

### LINKED ISSUE #193: Container resize column recalculation

wizarrc:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Fixes #193 and supersedes PR ~~#1034~~ **#1033** 

This adds the ability to detect table resizes, not just whole window resizes. Also simplifies the visibility observer directive.

**What is the new behavior?**

- [ ] Yes
- [x] No

**Other information**:


Hypercubed:
Interesting... did you consider using an existing Resize Observer poly-fill like https://github.com/que-etc/resize-observer-polyfill?  Something that will default to native when available?

wizarrc:
@Hypercubed nope, sure didn't, more like solving an issue I was having. I came up with the name myself too, so had no idea there was another one with the same name out there, lol

wizarrc:
@Hypercubed looks like `resize-observer-polyfill` has a 20ms delay, do you know if it uses timers? The one thing I liked about the approach I ported from was it's an actual event that triggers the callback instead of a timer every 20ms to check the width and height like what else I saw around the web.

As far as I know there is no need for native/non-native support since it uses a `scroll` event under the hood. Anyways, thanks for the feedback, looks like a pretty good library.

wizarrc:
@Hypercubed It does look like the library you mentioned is better in most cases, and is well tested. Would you like to bring that in as a dependency instead? I noticed there really isn't any dependencies other than the standard Angular and rxjs used in this project, thus why I made a simple port of a `scroll` based observer that fit into one file. I think my solution works better than currently, but I'm not aware of all the edge cases out there, and the library you mentioned might do a better job with legacy browsers. If that's what you'd prefer I can see if I can bring it in instead.

Hypercubed:
What I have been considering is creating a single resize observer directive, using the polyfill mentioned, that we could use across sever libs... including ngx-datatable.  I have't yet decided exactly how or if this should be done.  Feedback is welcome.

wizarrc:
@Hypercubed ah, makes sense. This seems like a big problem with style-able components, so a common angular-style directive makes sense, which is what I was trying to do, yet make it simple enough that it didn't require a dependency. I looked for an angular technique I could use, couldn't find any, so I resorted to something I could port simply and that seemed effective. I was not aware of all the prior work.

I can see if I can make a directive that adapts to this library that's generic for angular in general. I believe what I used only works with `div` elements. I was hoping for a simple fix this time around, but I've been getting push back on every step so far. I actually have a lot of ideas for things that can be factored out of ngx-datatable.

FYI, if you are interested, I am working on a fork of ngx-datatable with an Observable Collection model that doesn't require so much change detection pluming. I've switched to a service architecture internally, that helped me implement things such as a virtual scroll bars (helps with perceived performance by blocking the scroll till the next frame), virtualized horizontal scrolling, cell selection via a service for more reliable scroll movements that don't rely on hacky "find next dom element, focus it, and hope it triggers a browser scroll" semantics, a custom Angular Zone that handles single level deep events more efficiently without doing a full change detection for low latency events such as scrolling. A lot of this work was made specifically for the datatable, but I found it to be very generic in nature and it would be nice to extract it out into it's own library.

src/components/resize-observer.component.ts
```diff
@@ -0,0 +1,140 @@
...
+    private expandChild: HTMLElement;
+    private shrink: Element;
+    private resizeSensor: HTMLElement;
+
+    private _enabled: boolean = false;
+    private onResized: () => void;
+    private onScroll: () => void;
+    
+    constructor(element: ElementRef) {
+        this.resizeElement = element.nativeElement;
```
amcdnl: Tabs should be 2 spaces.

demo/basic/tabs.component.ts
```diff
@@ -58,11 +63,34 @@ import { Component } from '@angular/core';
       </div>
 
     </div>
-  `
+  `,
+  styles: [`
```
amcdnl: Styles should be inline.

amcdnl:
I'd prefer the polyfill too. I'm ok w/ that as a dependency, is it included in core-js though?

src/components/resize-observer.component.ts
wizarrc: okay

demo/basic/tabs.component.ts
wizarrc: @amcdnl inline how? I'm not sure I'm following

amcdnl: They should live in a separate file vs inline in the component.

wizarrc: thanks, will do

amcdnl:
I think we should close this PR in favor of the polyfill approach.

wizarrc:
@amcdnl Okay. I didn't see `MutationObserver` as part of the core-js polyfill.

wizarrc:
It seemed pretty easy to gut what I had and replace it with a polyfill based directive.

wizarrc:
> TS2307: Cannot find module 'resize-observer-polyfill'.

hmm, I thought adding to `peerDependencies` would be good enough. Not really sure why this is not working.

src/directives/visibility.directive.ts
```diff
@@ -14,50 +15,26 @@ import {
...
 
-  onVisibilityChange(): void {
+  onVisibilityChange(visible: boolean): void {
+    if(visible === this.isVisible) return;
+    this.isVisible = visible;
     // trigger zone recalc for columns
-    this.zone.run(() => {
-      this.isVisible = true;
-      this.visible.emit(true);
+    setTimeout(() => {
```
amcdnl: Why did you remove the zone run in favor of the settimeout?

src/directives/visibility.directive.ts
```diff
@@ -14,50 +15,26 @@ import {
  *
  */
 @Directive({ selector: '[visibilityObserver]' })
-export class VisibilityDirective implements OnInit, OnDestroy {
+export class VisibilityDirective {
```
amcdnl: Whats the point of this directive now you removed loop check?

src/directives/resize-observer.directive.ts
```diff
@@ -0,0 +1,60 @@
...
+  private resizeElement: Element;
+  private resizeObserver: ResizeObserver;
+  
+  private _enabled: boolean = false;
+  
+  constructor(element: ElementRef) {
+    this.resizeElement = element.nativeElement;
+
+    this.resizeObserver = new ResizeObserver((entries, observer) => {
+        for (const entry of entries) {
```
amcdnl: 2 space tabs please.

src/directives/visibility.directive.ts
wizarrc: It was removed because it was already running inside of the angular zone. setTimeout was needed because of timing issues.

src/directives/visibility.directive.ts
wizarrc: Not really sure it's needed anymore. It can be collapsed back into the main datatable.

alexw10:
@wizarrc Is this good to go? going to give this a try in our data table which is derived from ngx since we have this same issue with resize. This seems like a good way to solve it.

alexw10:
I am getting the following error ours is a little different where we are using rollup to bundle this all in the end:
'resize-observer-polyfill/index' is imported by dist/packages/vdl-angular/data-table/directives/resize-observer.directive.js, but could not be resolved – treating it as an external dependency
'resize-observer-polyfill/index' is imported by dist/bundles/vdl-angular.es5.js, but could not be resolved – treating it as an external dependency
No name was provided for external module 'resize-observer-polyfill/index' in options.globals – guessing 'ResizeObserver'

I do this in the TS like you do:
import ResizeObserver from 'resize-observer-polyfill';

Any chance you have any idea of how I would add this to our rollup globals? (I think this might be needed) otherwise any ideas at all?

src/directives/visibility.directive.ts
amcdnl: Does it work w/ the hidden examples w/o it?

src/directives/visibility.directive.ts
amcdnl: I really hate doing `setTimeout` for timing issues. Do you know what specifically is the issue?

amcdnl:
@alexw10 - Waiting on a few feedback items before I can merge.

wizarrc:
@alexw10 I haven't tried using rollup yet. I'll have to look into it. As for getting it merged, I need to remove the visibility directive and figure out some change detection issues before this is ready for merging.

RicardoVaranda:
Hey guys, any update on this?

amcdnl:
Can you resolve the conflicts?

wizarrc:
@RicardoVaranda sorry for the delay. I had to stop and think about the side effects of removing some code in terms of backwards compatibility. I had made the same changes in my fork and I had noticed all sorts of weird issues. I will try to get back to this PR in the next few days.

RicardoVaranda:
@wizarrc Don't be silly mate, thanks for all the work you're putting into this :)

src/directives/visibility.directive.ts
PMabMI: I was looking into this pull request and to answer your question, without the setTimeout call I'm getting the following error at runtime.

Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'enabled: undefined'. Current value: 'enabled: true'.

I'm thinking that using setTimeout to cause the emit to be run outside of the current call stack is pushing the change of the value to after this error check.

Issue #1064, opened 2017-10-19
Sorting has no effect when used with row grouping


idpaterson:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
Sorting specified with the `sorts` input or modified by clicking a column header is ignored when using `groupRowsBy`.

I attempted a workaround to null out `groupRowsBy` but while the group headers went away it did not fix the sort. In my case, it is possible to pre-sort the data before providing to `rows` so that groups are displayed in the correct order.

**Expected behavior**
Sorting should do *something* though the exact behavior requires discussion. I expected `sorts` to affect the order of groups and rows within those groups such that a primary sort on the group and secondary sort on the items in the group would result in sorted groups containing sorted items. 

However, I don't have a clear expectation for further sorting by the user. In my case it would be appropriate to either remove the groups once sorted on a different column, or maintain the group order but apply the selected sort within each group.

**Reproduction of the problem**
The [Row Grouping](http://swimlane.github.io/ngx-datatable/#row-grouping) example exhibits this behavior; clicking a column does not affect the data, it just re-expands all groups. 

![groupsort](https://user-images.githubusercontent.com/507058/31791370-5d3177ca-b4e6-11e7-9ba4-782962673274.gif)

**What is the motivation / use case for changing the behavior?**
It seems that all other features support sorting, so I chose to classify this as a bug.

**Please tell us about your environment:**
macOS, Atom, npm 5

* **Table version:** 10.3.0
* **Angular version:** Angular 4.0.3
* **Browser:** all
* **Language:** TypeScript 2.4 (required for ngx-datatable 10 and doesn't seem to break our Angular 4)

dariobraun:
Any chance this will be implemented? The absence of this feature is unfortunately a deal breaker.

MateusSilva0101:
Any solution for this case?

PR #1072, opened 2017-10-23
Allow sorting grouped table rows and reverting to ungrouped state

idpaterson:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

#1064 Sorting has no effect when used with row grouping

Summary:
- `sorts` and column header sorting do not change the order of grouped rows in any way
- Once grouped, table rows cannot be ungrouped

**What is the new behavior?**
Grouping is applied to the sorted `internalRows` rather than the raw `rows` input so any changes to sorting can affect the order of groups and of rows within groups. 

Setting `groupRowsBy` to `null` will remove all row grouping. Care has been taken to allow the `ngx-datatable-group-header` markup shown in examples to not throw errors when rows are not grouped.

Rows are sorted lazily via accessors, so rapid changes to `sorts` and `groupRowsBy` may not incur as much overhead since they now just invalidate the sorted rows. Once sorted and grouped the rows are cached for subsequent access until invalidated.

- [x] Yes
- [ ] No

Tables with grouped rows for which sorting should not be allowed will need to set `externalSorting` to `true` as rows will now automatically sort in response to user interaction with the column headers. Perhaps someone noticed and wanted this behavior.

Setting the `groupRowsBy` input to `null` will *immediately* remove any information about grouped rows, including anything set by the `groupedRows` input. Previously, it may have been possible to set both `groupedRows` and `groupRowsBy` then remove the latter, retaining the original `groupedRows`. However, this seems like a strange use case.

**Other information**:
This addresses the question of how sorting should be handled when rows are grouped in only the most basic fashion. Grouped rows are always derived from the result of sorting the rows according to `sorts`. Therefore, changing the sort order of columns can change the order of groups since group order is determined by the position of the first row exhibiting that group trait. 

A few alternate sort methods when rows are grouped:
- Set `groupRowsBy = null` to disable row grouping when sorting on other columns (this PR makes this possible) so that the sort applies to all rows without groups.
- Set `externalSorting` when sorting on other columns and sort rows with respect to the groups.
- Set `externalSorting = true` and ignore sorting events to disable sorting

### DIFF
--- a/src/components/body/body-row-wrapper.component.ts
+++ b/src/components/body/body-row-wrapper.component.ts
@@ -9,7 +9,7 @@ import { mouseEvent } from '../../events';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <div 
-      *ngIf="groupHeader && groupHeader.template"
+      *ngIf="groupHeader && groupHeader.template && row?.value"
       class="datatable-group-header"
       [ngStyle]="getGroupHeaderStyle()">
       <ng-template
@@ -20,7 +20,7 @@ import { mouseEvent } from '../../events';
     </div>
     <ng-content 
       *ngIf="(groupHeader && groupHeader.template && expanded) || 
-             (!groupHeader || !groupHeader.template)">
+             (!groupHeader || !groupHeader.template || !row?.value)">
     </ng-content>
     <div
       *ngIf="rowDetail && rowDetail.template && expanded"
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -672,7 +672,7 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
    * Returns if the row was expanded and set default row expansion when row expansion is empty
    */
   getRowExpanded(row: any): boolean {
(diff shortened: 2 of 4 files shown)


lautarobock:
Any news about it?

idpaterson:
@lautarobock, I had to move away from actual row grouping and just use a custom rowClass and sort to make certain records look like group headings. After working around the inability to sort I was faced with additional features that did not work quite as expected against grouped data. For example, paging applies to the number of groups on the page rather than the number of results and filtering results by a keyword can be extremely latent due to extra layout overhead when groups are enabled (in my case with non-debounced filtering the browser froze for about 800ms between each keystroke).

lautarobock:
I've created another PR trying to solve the merge conflicts

BenjaminBini:
hi guys, any news on this PR ?

Guling85:
Any news on this guys?

enavarro222:
Any news on this PR ?

MateusSilva0101:
> Alguma novidade sobre esse PR?

Any solution for this case?

Issue #817, opened 2017-06-14
Column custom sort: pass row data to comparator


PeS82:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[X] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
When TableColumn.comparator is used, arguments passed to the comparator are values of that column.

**Expected behavior**
It would be nice to have an option to receive row data to comparator.

Due to backward compatibility it would be nice to add another Input to TableColumn, e.g. rowComparator(row, row), not to break legacy code.

**What is the motivation / use case for changing the behavior?**

My specific case is I am displaying time (duration) in a column (e.g. "1d 20h 14m", "5h 20m"). In other column there is the same value, as number in seconds. Obviously, alphanumeric sorting is wrong in this case and I need to sort the column by seconds.

Since comparator feeds me with string value I must parse the duration back to seconds despite the fact that the value is already present in the table model.

I believe this would be handy in other cases when sortable value is already present however the displayed value is somehow "humanized".

* **Table version:** 
9.3.0

* **Angular version:** 

4.1.3 but that is not relevant I believe

olaf89:
I have the exactly the same issue, any solutions?

commitBlob:
same here

Ecksters:
Latest version has this feature, both rows are passed as the 3rd and 4th parameters to the comparator function, although I don't think it's been pushed to NPM yet.

chrisbmoore:
This should have been added in ngx-datatable 10.4.0. I've tried it and it works.

Ryan-Haines:
Does anyone have an example of providing a custom comparator in the template? Supposedly someone figured it out at the end of this thread: https://github.com/swimlane/ngx-datatable/issues/766 but the example is not complete and I have not been able to get it working.

ZackAfro:
Hi, hijacking a bit but I would like to know if you could add the comparator in a array object when you adding the columns using an array object and using the [columns]="columnsArray" input property. Something like this: 

`export const columns = [
  {
    prop: 'value',
    name: 'Heading',
    comparator: (a, b) => { return a - b }
  },`

arpansac:
@Ryan-Haines, thanks this worked perfectly.
Any way to use multiple comparators across separate columns? I've used it but it always binds to the last comparator.
Been stuck at it for some time now...

alexandis:
[...]

PR #1074, opened 2017-10-25
Row comparator

SPKorhonen:
- [ ] Bugfix
- [x ] Feature
- [ ] Build related changes
- [ ] CI related changes

Column comparators get only the value of current cell. This prevents more complex comparators which would need to access to the whole row. This problem has been referred to in issue #817.

**What is the new behavior?**
The rows being compared are added as the 3rd and 4th parameters to the comparator function. This preserves backward compatibility as js functions silently ignore additional parameters. This PR resolves issue #817 with minimal changes.

- [ ] Yes
- [x ] No

**Other information**:

### DIFF
--- a/src/utils/sort.ts
+++ b/src/utils/sort.ts
@@ -83,8 +83,8 @@ export function sortRows(rows: any[], columns: any[], dirs: SortPropDir[]): any[
       const propB = valueGetter(b, prop);
 
       const comparison = cachedDir.dir !== SortDirection.desc ?
-        cachedDir.compareFn(propA, propB, a, b) :
-        -cachedDir.compareFn(propA, propB, a, b);
+        cachedDir.compareFn(propA, propB, a, b, true) :
+        -cachedDir.compareFn(propA, propB, a, b, false);
 
       // Don't return 0 yet in case of needing to sort by next property
       if (comparison !== 0) return comparison;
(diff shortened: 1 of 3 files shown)


src/utils/sort.ts
```diff
@@ -83,8 +83,8 @@ export function sortRows(rows: any[], columns: any[], dirs: SortPropDir[]): any[
       const propB = valueGetter(b, prop);
 
       const comparison = cachedDir.dir !== SortDirection.desc ?
-        cachedDir.compareFn(propA, propB) :
-        -cachedDir.compareFn(propA, propB);
+        cachedDir.compareFn(propA, propB, a, b, true) :
```
amcdnl: Can you add a comment here whats going on. Also, better name for a/b would be good too.

SPKorhonen: I added some comments to the code and renamed a/b to rowA and rowB which should be less cryptic. Also I added few lines to documentation for consideration.

Confusedfish:
Thanks for this @SPKorhonen it solved my problem perfectly. I needed to place a total row into the table and didn't want to use the footer and without this it wasn't possible as the total row would jump about depending on the sorted column.

SPKorhonen:
I had a very similar use case, however I needed to bubble certain rows to the top regardless of sort. Luckily this functionality could be implemented without breaking too many things.

Confusedfish:
Where I lost most of the time was figuring out how to build the npm package. I am not sure if there is an issue with my environment or if it is windows specific but I could not reference the package via npm link once built on an aot project. I had to resort to building in a Linux docker and npm installing the resulting .gz file. What process do you use to build and what platform are you on?

Actually referencing your code took minutes :)

Ecksters:
Thank you so much for this, awesome feature.

amcdnl:
No problem. It’s in latest release

Issue #1120, opened 2017-11-14
Can't compile production build -aot


mtabaj:
I reopen the point, it seems that issue #1104 is not taken in account.
It absolutly blocks all the projects that use ngx-datatable and must use angular 5 

I'm submitting a ...

[X] **bug report**
Current behavior
As of 10.5.0, ngx-datatable triggers an error in zone.js subscribe - but only in a production build.

core.js:1350 ERROR Error: Uncaught (in promise): TypeError: Cannot read property 'subscribe' of undefined
TypeError: Cannot read property 'subscribe' of undefined
    at Pn (core.js:12014)
    at hr (core.js:13448)
    at Tr (core.js:13880)
    at Pr (core.js:13789)
    at hr (core.js:13476)
    at Tr (core.js:13880)
    at Pr (core.js:13789)
    at hr (core.js:13476)
    at cr (core.js:13338)
    at Object.Mr [as createRootView] (core.js:14025)
    at Pn (core.js:12014)
    at hr (core.js:13448)
    at Tr (core.js:13880)
    at Pr (core.js:13789)
    at hr (core.js:13476)
    at Tr (core.js:13880)
    at Pr (core.js:13789)
    at hr (core.js:13476)
    at cr (core.js:13338)
    at Object.Mr [as createRootView] (core.js:14025)
    at u (polyfills.76bc5925d9fb9c6d1bc0.bundle.js:1)
    at u (polyfills.76bc5925d9fb9c6d1bc0.bundle.js:1)
    at polyfills.76bc5925d9fb9c6d1bc0.bundle.js:1
    at t.invokeTask (polyfills.76bc5925d9fb9c6d1bc0.bundle.js:1)
    at Object.onInvokeTask (core.js:4617)
    at t.invokeTask (polyfills.76bc5925d9fb9c6d1bc0.bundle.js:1)
    at r.runTask (polyfills.76bc5925d9fb9c6d1bc0.bundle.js:1)
    at o (polyfills.76bc5925d9fb9c6d1bc0.bundle.js:1)
    at 
S @ core.js:1350
This error does not manifest in 10.4.0. It does not manifest when using a regular / debug build.

k20human:
Same problem here ...

villelahdenvuo:
For us using 11.0.4 it compiles just fine, but when loaded in the browser we get an error:

[...]

Issue #1114, opened 2017-11-08
10.5.0 Console error in production build


TomMettam:
**I'm submitting a ...** 
```
[X ] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
As of 10.5.0, ngx-datatable triggers an error in zone.js subscribe - but **only in a production build**.

<pre>core.js:1350 ERROR Error: Uncaught (in promise): TypeError: Cannot read property 'subscribe' of undefined
TypeError: Cannot read property 'subscribe' of undefined
    at Pn (core.js:12014)
    at hr (core.js:13448)
    at Tr (core.js:13880)
    at Pr (core.js:13789)
    at hr (core.js:13476)
    at Tr (core.js:13880)
    at Pr (core.js:13789)
    at hr (core.js:13476)
    at cr (core.js:13338)
    at Object.Mr [as createRootView] (core.js:14025)
    at Pn (core.js:12014)
    at hr (core.js:13448)
    at Tr (core.js:13880)
    at Pr (core.js:13789)
    at hr (core.js:13476)
    at Tr (core.js:13880)
    at Pr (core.js:13789)
    at hr (core.js:13476)
    at cr (core.js:13338)
    at Object.Mr [as createRootView] (core.js:14025)
    at u (polyfills.76bc5925d9fb9c6d1bc0.bundle.js:1)
    at u (polyfills.76bc5925d9fb9c6d1bc0.bundle.js:1)
    at polyfills.76bc5925d9fb9c6d1bc0.bundle.js:1
    at t.invokeTask (polyfills.76bc5925d9fb9c6d1bc0.bundle.js:1)
    at Object.onInvokeTask (core.js:4617)
    at t.invokeTask (polyfills.76bc5925d9fb9c6d1bc0.bundle.js:1)
    at r.runTask (polyfills.76bc5925d9fb9c6d1bc0.bundle.js:1)
    at o (polyfills.76bc5925d9fb9c6d1bc0.bundle.js:1)
    at <anonymous>
S @ core.js:1350</pre>

This error does not manifest in 10.4.0.

**Expected behavior**
I would expect the ngx-datatable component to behave as it does with a development build.

**Reproduction of the problem**

Package versions used:

"@angular/core": "^5.0.0",
"zone.js": "^0.8.4"
 "rxjs": "^5.5.2",
[...]

Issue #1104, opened 2017-11-06
Angular 5 issue


nitinht1988:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
After upgrading to angular 5 and also after updating version to ^11.0.0 i am getting below run time error when i try with  ng serve -aot, with ng serve it is working fine as expected

webpack-internal:///../../../core/esm5/core.js:1546 ERROR Error: Uncaught (in promise): TypeError: Cannot read property 'subscribe' of undefined
TypeError: Cannot read property 'subscribe' of undefined

**Expected behavior**
It should work without any error

**Reproduction of the problem**
update angular version to 5 and ngx-datatable version to  11.0.0 and run the angular project with ng serve -aot command.

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

Windows 7
visual studio code

* **Table version:** 0.8.x
 11.0.0

* **Angular version:** 2.0.x
5.0

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
Chrome latest

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
typescript 2.4.0

arlowhite:
Sounds like a RxJS/TypeScript error.

After upgrading to Angular 5 and angular-cli 1.5, I was not able to build until I updated TypeScript from 2.4 to 2.5

So try TypeScript 2.5.3 (latest) and see if it works.

arsyed04:
Latest Typescript is 2.6.1

adriantoczydlowski:
The same issue: **ERROR Error: Uncaught (in promise): TypeError: Cannot read property 'subscribe' of undefined. TypeError: Cannot read property 'subscribe' of undefined**
version: 11.0.0
Angular: 5.0.0
cli: 1.5.0
node: v8.7.0
typescript: 2.5.0

olaf89:
I have same error

amcdnl:
What is your rx version?

arsyed04:
I have opened issue with Angular since it seems more relevant there: https://github.com/angular/angular/issues/20240

solnat:
I had the same issue. 
[...]

PR #1121, opened 2017-11-14
Workaround AOT issue with Angular 5

Confusedfish:
- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

Using ngx-datatable on Angular 5 AOT fails in the browser with Cannot read property 'subscribe' of undefined. See #1104 #1120 possibly #1114

**What is the new behavior?**

AOT builds no longer error in browser.

- [ ] Yes
- [x] No

**Other information**:

The result of the tsconfig change will be that the compiled JS is not optimised to the extent that variables not set in the constructor are removed but presumably other optimisations will also be excluded. For example see \components\body\body.component.ts ln 216:

`rowHeightsCache: RowHeightCache = new RowHeightCache();` which isn't contained within the constructor. Technically perfectly valid but apparently stripped out by tsickle. See https://github.com/angular/angular/issues/19587 for more detail.

At some point in the future Angular will catch up with the version of TS and non constructor defaults will be once again supported at which point it would be best to revert this change.

### DIFF
--- a/tsconfig-aot.json
+++ b/tsconfig-aot.json
@@ -40,6 +40,7 @@
   "angularCompilerOptions": {
     "genDir": "release/",
     "strictMetadataEmit": true,
-    "skipTemplateCodegen": true
+    "skipTemplateCodegen": true,
+	"annotationsAs": "decorators"
   }
 }


amcdnl:
Releasing 11.0.4

villelahdenvuo:
We still have a problem using AOT on Angular 5 and ngx-datatable 11.0.4

image

Works fine without AOT.

PR #1181, opened 2017-12-08
Fixed the translated3D Z value for each row to support dropdown in a …

### LINKED ISSUE #1180: Transform translate3D blocks dropdown in a cell
```
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
If you try using a drop-down in any of the cells while **scrollbarV** is true and each row has a height the drop-down will be hidden under the next row.
An example is shown here - https://jsfiddle.net/sanborn/ejxregnc/

**Expected behavior**

The dropdown should be visible in that case. 
A solution example is shown here - https://jsfiddle.net/sanborn/ejxregnc/5/

Just try adding a drop-down in any of the cells.

We may need a drop-down menu for each row.

* **Table version:** 0.8.x

* **Angular version:** 2.0.x

sanbornsen:
…cell (fixes #1180)

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: Support for dropdowns in cell

https://github.com/swimlane/ngx-datatable/issues/1180

**What is the new behavior?**

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -29,6 +29,7 @@ import { MouseEvent } from '../../events';
         [scrollbarH]="scrollbarH"
         [scrollHeight]="scrollHeight"
         [scrollWidth]="columnGroupWidths?.total"
+        style="transform-style: preserve-3d"
         (scroll)="onBodyScroll($event)">
         <datatable-row-wrapper
           [groupedRows]="groupedRows"
@@ -455,8 +456,8 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
    * manipulated.   As an example, if the height of row 0 is 30 px and row 1 is
    * 100 px then following styles are generated:
    *
-   * transform: translate3d(0px, 0px, 0px);    ->  row0
-   * transform: translate3d(0px, 30px, 0px);   ->  row1
+   * transform: translate3d(0px, 0px, 2px);    ->  row0
+   * transform: translate3d(0px, 30px, 1px);   ->  row1
    * transform: translate3d(0px, 130px, 0px);  ->  row2
    *
    * Row heights have to be calculated based on the row heights cache as we wont
@@ -471,7 +472,6 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
    */
   getRowsStyles(rows: any): any {
     const styles = {};
-
     // only add styles for the group if there is a group
     if (this.groupedRows) {
       styles['width'] = this.columnGroupWidths.total;
@@ -493,9 +493,10 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
       // until the previous row position.
       const pos = this.rowHeightsCache.query(idx - 1);
 
-      translateXY(styles, 0, pos);
+      // Z position of translate3d
+      const zValue = this._rowCount - idx;
+      translateXY(styles, 0, pos, zValue);
     }
-
     return styles;
   }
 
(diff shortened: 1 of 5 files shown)


src/components/body/body.component.ts
```diff
@@ -29,6 +29,7 @@ import { MouseEvent } from '../../events';
         [scrollbarH]="scrollbarH"
         [scrollHeight]="scrollHeight"
         [scrollWidth]="columnGroupWidths?.total"
+        [ngStyle]="{'transform-style': 'preserve-3d'}"
```
amcdnl: No reason to make this `ngStyle`, just do a normal inline style.

src/components/body/body.component.ts
```diff
@@ -485,19 +485,20 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
...
-      }        
+      }
 
       // const pos = idx * rowHeight;
       // The position of this row would be the sum of all row heights
       // until the previous row position.
       const pos = this.rowHeightsCache.query(idx - 1);
 
-      translateXY(styles, 0, pos);
+      // Z position of trnaslate3d
```
wizarrc: spelling error

wizarrc:
@sanbornsen This looks like in interesting fix. Can you provide a demo of how this feature works? Does this work with dropdown menus at the bottom of the view? This looks like it has a direction bias for an overlay menu.

In my scenario, say you want to right click which triggers a dropdown menu, but the menu would clip past the view. In many desktop programs, the dropdown menu would be responsive to the constraint in instead display starting at the top of the row instead of the bottom of the row when there is no space available.

amcdnl:
I tend to agree w/ @wizarrc - typically you want to move your dropdown to the body and then position relatively for this very reason. Checkout - https://medium.com/@amcdnl/building-tooltips-for-angular2-396320fa938f

wizarrc:
@amcdnl without a demo, I am skeptical this would even work if a drop down menu is implemented inside a row or cell template. My understanding of how clipping works is because it's inside the row element, it can't escape past the cell, despite the  `z-index`. I didn't know this restriction can be bypassed with `translate3d` css. A demo would help test my assumptions.

sanbornsen:
I have added a basic demo. 

True that this solution will not work for dropup but only for dropdown.

src/utils/translate.ts
```diff
@@ -9,13 +9,13 @@ const hasCSS3DTransforms = typeof window !== 'undefined' ? !!getVendorPrefixedNa
...
     if (!isSafari && hasCSS3DTransforms) {
-      styles[transform] = `translate3d(${x}px, ${y}px, 0)`;
+      styles[transform] = `translate3d(${x}px, ${y}px, ${z}px)`;
       styles[backfaceVisibility] = 'hidden';
     } else {
-      styles[camelCase(transform)] = `translate(${x}px, ${y}px)`;
+      styles[camelCase(transform)] = `translate(${x}px, ${y}px, ${z}px)`;
     }
   } else {
     styles.top = `${y}px`;
```
amcdnl: What about support for top/left scenarios?

sanbornsen: So the drop-down components usually have its own high z-index value. Therefore, it should work for top/left scenarios.

sanbornsen:
We are also adding the same feature for the header in a more elegant way. So hold on to this PR.

PR #1197, opened 2017-12-14
Virtual row works with state change

sanbornsen:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Any event from a cell that changes a state of any row doesn't work on virtually rendered rows. 

**What is the new behavior?**
Even from virtually rendered rows any even can change a state of a row and it reflects.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/components/body/body-row.component.ts
+++ b/src/components/body/body-row.component.ts
@@ -1,5 +1,5 @@
 import {
-  Component, Input, HostBinding, ElementRef, Output, KeyValueDiffers, KeyValueDiffer,
+  Component, Input, HostBinding, ElementRef, Output, KeyValueDiffers, KeyValueDiffer, NgZone,
   EventEmitter, HostListener, ChangeDetectionStrategy, ChangeDetectorRef, DoCheck, SkipSelf
 } from '@angular/core';
 
@@ -30,7 +30,7 @@ import { MouseEvent, KeyboardEvent, Event } from '../../events';
         [displayCheck]="displayCheck"
         (activate)="onActivate($event, ii)">
       </datatable-body-cell>
-    </div>      
+    </div>
   `
 })
 export class DataTableBodyRowComponent implements DoCheck {
@@ -47,12 +47,12 @@ export class DataTableBodyRowComponent implements DoCheck {
   @Input() set innerWidth(val: number) {
     if (this._columns) {
       const colByPin = columnsByPin(this._columns);
-      this._columnGroupWidths = columnGroupWidths(colByPin, colByPin);  
+      this._columnGroupWidths = columnGroupWidths(colByPin, colByPin);
     }
 
     this._innerWidth = val;
     this.recalculateColumns();
-    this.buildStylesByGroup();    
+    this.buildStylesByGroup();
   }
 
   get innerWidth(): number {
@@ -61,7 +61,12 @@ export class DataTableBodyRowComponent implements DoCheck {
 
   @Input() expanded: boolean;
   @Input() rowClass: any;
-  @Input() row: any;
+  public row: any;
+  @Input('row') set rowSetter(val: any) {
+    this.zone.run(() => {
+      this.row = val;
+    });
+  }
   @Input() group: any;
   @Input() isSelected: boolean;
   @Input() rowIndex: number;
@@ -123,8 +128,9 @@ export class DataTableBodyRowComponent implements DoCheck {
   constructor(
       private differs: KeyValueDiffers,
       @SkipSelf() private scrollbarHelper: ScrollbarHelper,
-      private cd: ChangeDetectorRef, 
-      element: ElementRef) {
+      private cd: ChangeDetectorRef,
+      element: ElementRef,
+      private zone: NgZone) {
     this._element = element.nativeElement;
     this._rowDiffer = differs.find({}).create();
   }
@@ -134,7 +140,7 @@ export class DataTableBodyRowComponent implements DoCheck {
       this.cd.markForCheck();
     }
   }
-  
+
   trackByGroups(index: number, colGroup: any): any {
     return colGroup.type;
   }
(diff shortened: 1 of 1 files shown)


src/components/body/body-row.component.ts
```diff
@@ -61,7 +61,12 @@ export class DataTableBodyRowComponent implements DoCheck {
 
   @Input() expanded: boolean;
   @Input() rowClass: any;
-  @Input() row: any;
+  public row: any;
+  @Input('row') set rowSetter(val: any) {
```
amcdnl: Lets name this `row` and make `this.row` be `this._row`

src/components/body/body-row.component.ts
```diff
@@ -61,7 +61,12 @@ export class DataTableBodyRowComponent implements DoCheck {
...
+    this.zone.run(() => {
```
amcdnl: Wouldn't `cd.markForCheck` be better here?

sanbornsen: I don't know much about **ChangeDetectorRef** , I'll try that out, if works I'll change it.

src/components/body/body-row.component.ts
```diff
@@ -61,7 +61,11 @@ export class DataTableBodyRowComponent implements DoCheck {
 
   @Input() expanded: boolean;
   @Input() rowClass: any;
-  @Input() row: any;
+  @Input() set row(val: any) {
+    this.zone.run(() => {
```
sanbornsen: **this.cd.markForCheck();** seems to not working here.

wizarrc: @sanbornsen `markForCheck` only marks the component to be checked for change. It relies on `detectChanges` to be run from a parent component before the changes are performed. Angular hooks into `zone.run` and calls `detectChanges` on the root element after the method completes.

I would suggest you check why change detection isn't being performed after the change has been set. The reason is most likely because the component is run outside of the Angular zone `NgZone`.

Can you provicde a test case where this fails necessitating the need for this PR?

sanbornsen: @wizarrc let me try to explain the case. Assume there is a button in each row with a click event that changes a state in the host component. For the initially loaded set of rows, that works fine. As soon as you scroll and try the same click on some virtually rendered row, the state changes but changeDetection fail to acknowledge the change.

For example, run this branch - https://github.com/sanbornsen/ngx-datatable/tree/virtual_row_test and go to http://localhost:9999/#virtual-scroll

sanbornsen: I honestly couldn't figure out, why is it not running in the zone. Please explain if you understand the problem. This is very interesting !!

wizarrc: @sanbornsen Ah, I had this same problem in my fork. What really is needed in Angular is a creation zone context. Here is the problem: When you listen to a scroll event, it is listened outside of the angular zone for performance reasons. The problem arises when the virtualizer kicks in and materializes the rows and cells and recycles the old ones. A side effect of that results in all templates being rendered and new events handled outside of the angular zone.

What I did to solve that was to create a custom zone service that listens to the scroll event, and intercept all calls to datatable events. After handling the events, the service determines which events require change detection based on if anyone is listening to the events outside the library. It wasn't that complicated of a fix, but I already had designed a zone service prior, so all I had to do was design a creation context concept to my service. Fixing it on here would be more challenging.

sanbornsen: @wizarrc Thanks for the explanation. It makes sense now. So, what I am doing is, I am putting the template rendering back to the zone and thus it's working. Looks like a solution to me. 

Can you share your zone service code, if it's opensource? It sounds like creating a zone service listening to the scroll is violating the purpose of keeping the scroll out of the Angular zone.

wizarrc: > Can you share your zone service code, if it's opensource?

@sanbornsen yeah, when I get around to it. I really needed a performant component for a very specific use case. I broke a lot of stuff trying to add perf, but I'm slowly adding back broken features. I'm in the process of stabilizing.

> It sounds like creating a zone service listening to the scroll is violating the purpose of keeping the scroll out of the Angular zone.

That would be true if it were to cause a full Angular refresh like the Angular zone does with it's hooks. My custom zone forks from outside the Angular zone and it's hook only refreshes the Datatable component instead of the whole world after various events like scrolling. If no one subscribes to the event outside of the datatable, I simply ignore it, however, if someone subscribes to the event, it will trigger full change detection. Well not quite, I do a trick in there first to prevent full change detection based on prorogation of the event by monkey patching the event processing, but I can only do that one parent level down unfortunately.

amcdnl: What about calling `detectChanges` instead of `markForCheck`?

wizarrc: @sanbornsen for what it's worth, looking back I really wish I could have made things more modular in my fork so I could piecemeal changes back into the original project. It's been really hard to do so, mostly from the lack of testing coverage. I break way too many things without knowing till I do manual testing. I really need much better acceptance tests before I open it up to everyone.

I noticed that here with this project too. It feels like people using this library are stuck in a perpetual version fragmentation. Someone adds a cool new feature, and the maintainers look over the code and it seems okay, but it breaks people in weird obscure ways. I'm all for new features, but I really feel the focus should be on testing the current code, to weed out problems like this that could have been caught in a unit test. In many cases, a single fix will cause a regression in another area. In this case, it would cause a performance regression. To fix the performance regression, new code would be added to give a so-called "creation context" for handling events on materialized templates. Adding that might cause a regression bug somewhere else. Without good test coverage, I can't be confident my fix isn't hurting anyone else.

@amcdnl Respectfully I suggest that in order for contributors to check in new code, it must come with test code pointing out the problem for the bug fix. If they are unable to provide it, then either you are myself create one for the fix before merging the PR. I am still fairly new to unit testing but I'm really starting to see it's value. I'm struggling myself from regressions in my own fork, and feel like it's wasted time that could have been prevented.

wizarrc: > Suppose that template run outside of the angular zone changed more than just row but also rowClass, expanded among others in the same cycle.

What I meant is that I'm sure I could file a bug report for any property not being updated after a scroll event due to materialization on a non-tracked zone. `sanbornsen` noticed a bug for his use case, but I'm sure this is a more general problem.

PR #1225, opened 2018-01-09
Add resize handler for non-resizable columns

artem-palamarchuk96:
- [ ] Bugfix
- [x] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

There is no resize handler element when column is not resizable, so we cannot apply the same styles like, fox example, border.

**What is the new behavior?**
Now there is also exists resize handler span element when column is no resizable. It has different CSS class so we can put some different styles to it

- [ ] Yes
- [x] No

**Other information**:


amcdnl:
I'm a bit confused on the purpose of this. You want to add a handle when there is no resize? Seems odd.

artem-palamarchuk96:
There is next situation http://joxi.ru/v29nlnKiGDaOKA first column is not re-sizable and other are re-sizable, fro re-sizable columns we have span tag with "resize-handle" css class, and i can apply for example style of our border between column such as white color, and when column is not re sizable we didn't have this span tag element and i need to make border in another way, and this is not consistent, i hope you understand what i'm talking about

amcdnl:
Gotcha. Accepted! Thanks for your contribution!

Issue #880, opened 2017-07-31
ERROR Error: query at index -1 failed: Fenwick tree array not initialized


ankita199:
I am Passing rowheight from external component and scrollbarV as a true.
![rowheight](https://user-images.githubusercontent.com/9448677/28764730-aa19f0f8-75e4-11e7-932c-875bd6d7df11.png)

amcdnl:
Can you make a demo?

waeljammal:
Same issue here, happens when I start with an empty list with server side virtual scrolling then I add a new record so there is 1 in the list now. Then delete that 1 record and it throws that error.

wizarrc:
@waeljammal are you using `externalPaging` and `count` when the error occurs?

waeljammal:
@wizarrc Yes I am, below are the fields I'm using for paging.

        [externalPaging]="true"
        [count]="repos.totalElements"
        [offset]="repos.pageNumber"
        [limit]="repos.size"

waeljammal:
As an experiment I tried changing the code to:

```
if (!this.treeArray.length) {
     return 0;
     //throw new Error("query at index " + atIndex + " failed: Fenwick tree array not initialized.");
}
```
And it seems to work fine, I'm not sure what the negative effect of such a change is, I'm not familiar with the code. Not yet anyway.

wizarrc:
@waeljammal you don't want to make such a change. that error message is more like an internal error, and should never happen.

wizarrc:
@waeljammal when you deleted the row, how did you make the change. The latest version is push based and requires a copy for ever update.

waeljammal:
@wizarrc understood, however is there an alternative fix I could implement in the interim?

wizarrc:
@waeljammal for example, instead of doing `this.rows.splice(0,1)`, do `this.rows = this.rows.slice(1)`, or after mutating the object do `this.rows = [...this.rows]`. Something like that.

waeljammal:
@wizarrc I make my changes like so:

    this.findAll('plugin-registries', offset, this.size).subscribe((repos: any) => {
      if (repos) {
        this.page = repos['page'];
        this.links = repos['_links'];

        // calc start
        const start = this.pageNumber * this.size;

        // copy rows
        const rows = [...this.data];

        // insert rows into new position
        rows.splice(start, 0, ...repos['_embedded'][this.serviceName]);

        // set rows to our new rows
        this.data = rows;
      }
    });

[...]

PR #1229, opened 2018-01-11
#880, make the demo at /#virtual-paging work properly

ivankolev:
set rowCount to rows.length, to prevent 'Error: query at index -1 failed: Fenwick tree array not initialized.' and make the demo work properly at https://swimlane.github.io/ngx-datatable/#virtual-paging

- [ x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

As described in the issue #880 and currently on https://swimlane.github.io/ngx-datatable/#virtual-paging or http://localhost:9999/#virtual-paging, virtual paging demo throws an error because the rowCount does not properly reflect the number of rows

**What is the new behavior?**
Fixes the said javascript error and make the demo work as intended

- [ ] Yes
- [x ] No


src/components/body/body.component.ts
```diff
@@ -125,6 +125,7 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
   @Input() set rows(val: any[]) {
     this._rows = val;
     this.rowExpansions.clear();
+    this._rowCount = this._rows.length;
```
amcdnl: This won't work right since the row count varies. There is a function to get this information.

ivankolev:
Ok, back to the drawing board then. It fixes the error and all the tests passed, so I assumed it was enough. So, just to make it clear, there could be cases where rowCount is not equal to the number of rows?

amcdnl:
Server side paging.

src/components/body/body.component.ts
ivankolev: Aah, I see that /#server-paging is indeed broken with my change. Wonder if it's worth it to maybe add more e2e tests that could potentially catch such errors.

ivankolev:
Alright, I have pushed my second attempt at it, both /#server-paging and /#virtual-paging seem to be working now...

Val1962:
@amcdnl Any directions if you think suggested fix is not good?
It would be good to make virtual-paging working, the first page used to load successfully. I think it stopped working after scroller performance fix https://github.com/swimlane/ngx-datatable/commit/7227e559107f531fdeaf056d9c79133dcb62fdbf to run 'scroll' listener outside of Angular.

Issue #1232, opened 2018-01-11
Route change event from table row click disrupts Angular change detection on Edge


hoffination:
**I'm submitting a ...**  (check one with "x")
```
[ x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
I have a ngx-datatable that I have setup with each row to act like a button to route to another view. On click, a route change is triggered and the view is replace by another page which should destroy the page the table is on. This seems to work great on Chrome but I'm having issues on Edge.

When Edge makes the change there is a bug where the following view isn't rendered unless I trigger change detection manually with `ChangeDetectorRef.detectChanges()`. The routed view seems to only render the template with little-to-no bootstrapping of the component. The original page is still visible below the new view.

When I debug the problem it looks like the `constructor` is called but not the `ngOnInit` method. My current work-around is to run a `setTimeout` in the constructor and run change detection.

Another layer of strangeness is I'm only getting this bug on clicks from the second or later pages of content in the datatable. Clicks in the first rendered page seem to be fine and bootstrap the new route correctly. I'm also not getting any errors in the Edge console.

**Expected behavior**

Upon clicking on a row in the datatable the click handler would change the route and the following view will bootstrap correctly. The table will allow change detection to trigger regularly and process the route change appropriately.

**Reproduction of the problem**

I'll try to reproduce in a plnkr but here's an example of my current table use:

```
<ngx-datatable
    #datatable
    class="material striped"
    [headerHeight]="50"
    [footerHeight]="50"
    [rows]="outputRows$ | async"
    [selected]="selected"
    [selectionType]="'checkbox'"
    (select)='onSelect($event)'
    columnMode="flex"
    [selectAllRowsOnPage]="false"
    [rowHeight]="50"
    [scrollbarV]="true"
    (activate)="handleEvent($event)"
  >
    <ng-content></ng-content>
  </ngx-datatable>
```

**What is the motivation / use case for changing the behavior?**

Seems to be a serious bug on Edge.

**Please tell us about your environment:**

Basic Angular CLI built project served up on OSX and run on Browserstack's Edge 16 browser. Also reproduced on Windows 10 running Edge 16.

* **Table version:** 11.1.5

* **Angular version:** 5.2.0

* **Browser:** [ Edge 16 ]

* **Language:** [TypeScript 2.4.2]

hoffination:
[...]

PR #1238, opened 2018-01-15
Fix for #1232 causing cascading change detection disruption

hoffination:
added ngZone.run call to trigger change detection

- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

**What is the current behaviour?** (You can also link to an open issue here)
Change detection is not correctly firing when clicking on a row navigated to a new route, destroying the table. Some elements of the following view would not receive proper change detection. Triggering change detection manually would update the view, but was not triggered through the normal Angular lifecycle. This led us to believe there was an issue with zone.js.

This would only occur when routing from a virtualized row that had been rendered outside of the initial batch visible to the table. Example (render 10 visible rows, scroll down to 11th element and click, problems would occur). It would also occur when scrolling down and back up and clicking on one of the original elements. This led to the problem which was found in the ScrollerComponent: https://github.com/swimlane/ngx-datatable/blob/master/src/components/body/scroller.component.ts#L72

**What is the new behavior?**
Now change detection fires correctly on the subsequent route and the problem seems to have been resolved. Tests still pass and the application does not seem to have any slow down as a result.

- [ ] Yes
- [ x] No

**Other information**:
Shoutout to @bennett000 and @andrewthauer for helping me find a fix to the problem!


src/components/body/scroller.component.ts
```diff
@@ -69,7 +69,7 @@ export class ScrollerComponent implements OnInit, OnDestroy {
     requestAnimationFrame(() => {
```
bennett000: @hoffination I know this `requestAnimationFrame` was here before, but I'm curious as to why?  It might be more advantageous to debounce this but it seems out of scope for the PR.  I wonder if the authors had an intention that isn't obvious to me?

hoffination: Agreed. I'm not sure of the benefit

amcdnl:
Thats a crazy case. Thanks for the detailed explanation.

However, I purposely don't want to put that calculation inside the zone. The zone kills performance, I want to take control of the whole lifecycle and CD manually for the scrolling case.

Just to be clear, this is just when you are clicking on a router link inside a cell?

src/components/body/scroller.component.ts
amcdnl: See: https://www.html5rocks.com/en/tutorials/speed/rendering/. The rows are positioned using css animations.

bennett000: That makes sense, I usually debounce and I wondered if it was a vsync thing.

NauxLiu:
@amcdnl I ran into a similar problem, listening to the `activate dblclick` event on the virtualized row that triggered the pop-up dialog box, which resulted in an exception.

hoffination:
@amcdnl That makes sense. It happens when the `activate` callback is called on a row where we change the angular route from a component function. We filter the events down to just the click events. Not seeing any exceptions when this occurs.

amcdnl:
So if thats the case, why not put it in the activate callback?

hoffination:
@amcdnl Good suggestion! It looks like that resolved my issue in my project. I'd be happy to update the docs with a troubleshooting section including this solution if you can point me to the right place

hoffination:
Also closing the PR as this isn't the solution we're moving forward with.

PR #1239, opened 2018-01-16
Adding in explicit injections for a hybrid angular app

timruhle:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When running an [Angular Hybrid](https://github.com/ui-router/angular-hybrid/) app with both AngularJS 1.x and Angular 5 I get the error 

```Can't resolve all parameters for DataTableFooterTemplateDirective: (?)```

**What is the new behavior?**

Added in explicit injection so that angular can resolve all the parameters

- [ ] Yes
- [x] No

**Other information**:

This shouldn't change or affect anything, so there is no need to change any documentation or tests.

### DIFF
--- a/src/components/body/body-cell.component.ts
+++ b/src/components/body/body-cell.component.ts
@@ -1,7 +1,7 @@
 import {
   Component, Input, PipeTransform, HostBinding, ViewChild, ChangeDetectorRef,
   Output, EventEmitter, HostListener, ElementRef, ViewContainerRef, OnDestroy, DoCheck,
-  ChangeDetectionStrategy
+  ChangeDetectionStrategy, Inject
 } from '@angular/core';
 
 import { Keys } from '../../utils';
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

**Comment by amcdnl** [comment]
This shouldn't be the case. Can you make a demo showing what you are trying to do?

**Comment by timruhle** [comment]
@amcdnl I am using angular-hybrid to migrate and existing app, so I am not using angular-cli.

I have ```System.config({paths: {'@swimlane/ngx-datatable': '/node_modules/@swimlane/ngx-datatable/release/index.js'}});```

And then importing the library

```js
import { NgxDatatableModule } from '@swimlane/ngx-datatable';


@NgModule({
  imports: [
    NgxDatatableModule,
  ]
})
export class MyTableModule {}
```
And that's when it throws the error

```Can't resolve all parameters for DataTableFooterTemplateDirective: (?)```.


I haven't had this problem using angular cli and a pure angular 5 app, not with any other library. Unless there is something about angular dependency inject that I do not know about, I can't find any other solution.

Thanks for your help

PR #1244, opened 2018-01-17
Add setStylesByGroup() in set column prop

darkurse:
- [x] Bugfix

Column headers are not being resized on initialization
image

Without any changes to the code, only a windows resize seems to resolve this issue.

**What is the new behavior?**
It re-calculates the style (width) of each groups "left", "center", "right" after the columns have been set via the column property in `header.component.ts`.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/components/header/header.component.ts
+++ b/src/components/header/header.component.ts
@@ -101,6 +101,7 @@ export class DataTableHeaderComponent {
     const colsByPin = columnsByPin(val);
     this._columnsByPin = columnsByPinArr(val);
     this._columnGroupWidths = columnGroupWidths(colsByPin, val);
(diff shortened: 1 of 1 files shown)


stewx:
This change appears to have introduced a new defect: https://github.com/swimlane/ngx-datatable/issues/1319

It is not immediately obvious why, but using Git bisect, I narrowed it down to this commit.

stewx:
Can either of you offer insight? @amcdnl @darkurse

stewx:
PR to fix the defect introduced here: https://github.com/swimlane/ngx-datatable/pull/1326

This PR should have updated the logic for the body at the same time as it did for the header.

PR #1255, opened 2018-01-24
allowed observable in PipeTransform configuration of TableColumn options

souzara:
- [ ] Bugfix
- [x ] Feature
- [ ] Build related changes
- [ ] CI related changes

 Added optional feature, if a pipe that returns a Observable in transform method, it's will be subscribed and set value that the observable returns.

**What is the new behavior?**
Async pipe or Observables transformations in Pipe

- [ ] Yes
- [x ] No

**Other information**:

### DIFF
--- a/src/components/body/body-cell.component.ts
+++ b/src/components/body/body-cell.component.ts
@@ -8,6 +8,7 @@ import { Keys } from '../../utils';
 import { SortDirection } from '../../types';
 import { TableColumn } from '../../types/table-column.type';
 import { MouseEvent, KeyboardEvent } from '../../events';
+import { Observable } from 'rxjs/Observable';
 
 @Component({
   selector: 'datatable-body-cell',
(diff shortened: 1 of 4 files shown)


src/components/body/body-cell.component.ts
```diff
@@ -225,13 +226,21 @@ export class DataTableBodyCellComponent implements DoCheck, OnDestroy {
       const userPipe: PipeTransform = this.column.pipe;
 
       if (userPipe) {
-        value = userPipe.transform(val);
+        let transformResult = userPipe.transform(val);
+        if (transformResult instanceof Observable) {
+          let observable: Observable<any> = transformResult;
+          observable.subscribe((transformedValue) => {
```
amcdnl: You never unsubscribe from this, I think this would create a memory leak.

souzara: What do you think about something like this:

let pipeSubscription = observable.subscribe((transformedValue) => {
      value = transformedValue;
      pipeSubscription.unsubscribe();
 });

PR #1287, opened 2018-02-07
Changed page size calculation to count only fully visible rows

dlorych:
- [x] Bugfix
- [] Feature
- [ ] Build related changes
- [ ] CI related changes

Paging
-------
1. Calculation of the page size includes not fully visible row. If for example one pixel of a row is visible on a first page at the bottom, the next page would not contain it - it would be skipped.
2. Calculation of the page size does not take into account horizontal scrollbar height in a calculation. If for example a row is hidden behind a scrollbar, it will not be a part of the next page.
3. The minimum page size is 0, when the viewport is smaller than a row, paging is not available.

**What is the new behavior?**
Paging
-------
1. Calculation of the page size includes fully visible rows. If for example one pixel of a row is visible on a first page at the bottom, the row would be displayed as a first row on the next page. It would be also on the next page, if one pixel was missing to fully fit on the first page (!).
2. Calculation of the page size takes into account horizontal scrollbar height in a calculation. If for example a row is hidden behind a scrollbar, it will be shown as the first row on the next page.
3. The minimum page size is 1. When the viewport is smaller than a row, paging is still available - user change pages, which will show top parts of the rows.

- [ ] Yes
- [x] No

**Other information**:
Paging behavior is being changed here, so if someone did rely on previous behavior might be surprised after the change.

### DIFF
--- a/src/components/datatable.component.ts
+++ b/src/components/datatable.component.ts
@@ -896,8 +896,20 @@ export class DatatableComponent implements OnInit, DoCheck, AfterViewInit {
     // This is because an expanded row is still considered to be a child of
     // the original row.  Hence calculation would use rowHeight only.
     if (this.scrollbarV) {
-      const size = Math.ceil(this.bodyHeight / this.rowHeight);
-      return Math.max(size, 0);
+        var height = this.bodyHeight;
+        if (this.scrollbarH) {
+          // Make sure height does not include horizontal scrollbar if visible
+          let bodyEl = this.element.querySelector('.datatable-body')
+          height = bodyEl.clientHeight;
+        }
+        // Page size should include only fully visible rows.
+        // If one row is not fitting fully in a given height, 
+        // it should be a first row on the next page.
+        var size = Math.floor(height / this.rowHeight);
+        
+        // Minimum page size is 1, in case when viewport is smaller than row's height. 
+        // To still make possible changing pages
+        return Math.max(size, 1);
     }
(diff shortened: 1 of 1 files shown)


src/components/datatable.component.ts
```diff
@@ -896,8 +896,20 @@ export class DatatableComponent implements OnInit, DoCheck, AfterViewInit {
     // This is because an expanded row is still considered to be a child of
     // the original row.  Hence calculation would use rowHeight only.
     if (this.scrollbarV) {
-      const size = Math.ceil(this.bodyHeight / this.rowHeight);
-      return Math.max(size, 0);
+        var height = this.bodyHeight;
+        if (this.scrollbarH) {
+          // Make sure height does not include horizontal scrollbar if visible
+          let bodyEl = this.element.querySelector('.datatable-body')
```
amcdnl: Can this be cached?

dlorych:
I'm not sure whether caching it on a class property level is a way to go. Anyway, check and let me know what you think.

dlorych:
There was a problem with bodyEl.clientHeight. In case of window maximize (resize) the bodyEl.clientHeight was not yet calculated with new dimensions, which caused some problems.
Finally, I got rid of it in exchange for calculation whether scrollbarH is visible or not - based on column widths.
I'm not sure, though, whether column widht is always available... I have tested it only of 'fixed' columns.

PR #1308, opened 2018-02-22
Fix Default Sort Behavior Of Table With Custom Templates

swseverance:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

There are scenarios for which a table's `_internalColumns` object appears to be undefined when the initial sorting of the table occurs as a result of component lifecycle events. This can lead to the table not being sorted even when `sorts` are passed to the table via an input.

**What is the new behavior?**
Table rows are now sorted after changes to columnTemplates

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/src/components/datatable.component.spec.ts
+++ b/src/components/datatable.component.spec.ts
@@ -10,11 +10,11 @@ import {
 } from '.';
 import { NgxDatatableModule } from '../datatable.module';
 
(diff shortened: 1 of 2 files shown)


marjan-georgiev:
Looks good, thanks! 👍

Issue #1348, opened 2018-04-03
Dialog not working after scrolling 


duyleduc:
Bug, feature request, or proposal:
Bug

What is the expected behavior?
Page rendered correctly with passing data in parameter when clicked in a row after scrolled

What is the current behavior?
A blank page displaying after scrolled,  if I resize the windows in chrome debug console, it will be rendered.
I don't have problem if passing null value in parameter, happen only if there is an object passing in parameter.
This is the page when openned the dialog (unexpected behavior)
dialog openned
<img width="1092" alt="screen shot 2018-04-03 at 16 47 30" src="https://user-images.githubusercontent.com/6185711/38259259-dd79b108-3764-11e8-9874-12837633ad57.png">
After resizing the windows in chrome debug tool (expected behavior)
after resized
<img width="1027" alt="screen shot 2018-04-03 at 16 47 44" src="https://user-images.githubusercontent.com/6185711/38259260-dd9080fe-3764-11e8-9ac0-25ddcf732830.png">
Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Angular 5.0.0, material 5.2.2, OS: windows, mac os
Browser: Chrome

duyleduc:
Anyone could help me please?

duyleduc:
Issue resolved after wrapping the code of openning a dialog in zone.run()

deeg:
If anyone is interested to use a version with scrolling running in zone, you can replace your package.json:

"@swimlane/ngx-datatable": "semanticresearch/ngx-datatable#scroll-in-zone",

The only difference from release v13 is making scrolling happen inside zone:

https://github.com/semanticresearch/ngx-datatable/commits/scroll-in-zone

deeg:
@marjan-georgiev, I believe this should be fixed and can be closed with #1375 which was just merged.

Issue #1321, opened 2018-03-05
Angular Material Tooltip stop working after virtual scrolling


sheiidan:
**I'm submitting a ...**  (check one with "x")
```
[ x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
When one of the datatable cells contains an Angular Material2 Tooltip component, it will stop working after the user scrolls (the datatable is configured with vertical virtual scrolling). When this happens, you can see a blank tooltip shows at the top left corner of the page.

**Expected behavior**
Material Tooltip should always work regardless the user scrolls or not

**Reproduction of the problem**
1. Browse https://stackblitz.com/edit/angular-mq8vbn
2. move the cursor over the 1st column cell values, you should see 'Hello World' tooltips
3. now scroll down several rows
4. move the cursor over the 1st column cell values, this time the tooltip won't work.

(I suspect it might be a zone issues)

**What is the motivation / use case for changing the behavior?**
It breaks the tooltip

**Please tell us about your environment:**
N/A

* **Table version:** latest

* **Angular version:** 5.0.0

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

sheiidan:
Any update on this issue? It also happens when opening a material dialog after virtual scrolling. The dialog won't open unless you wrap it with zone.run()

GITHuHo:
We experience a similar issue with the ng-bootstrap tooltip if used with a template inside a cell. 
The template's data-bindings get lost after scrolling the table rows in and out of view.

Before scrolling...
![github_issue_ngxdatatable_1321_1](https://user-images.githubusercontent.com/9657001/37771157-0bb5b0e0-2dd7-11e8-8771-148cbbe98cb7.png)

After scrolling...
![github_issue_ngxdatatable_1321_2](https://user-images.githubusercontent.com/9657001/37771170-1186007e-2dd7-11e8-82ff-d3dda58a43dc.png)

```
<ng-template let-value="value" ngx-datatable-cell-template>
    <ng-template #valueToolTip>
        <span>The value is {{value}}</span>
    </ng-template>
    <div [ngbTooltip]="valueToolTip">
[...]

Issue #1223, opened 2018-01-06
Virtual scroll causes issues with auxilary (named) routes


yusijs:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
When virtual scroll is activated on a datatable, this breaks auxilary (named) routing. 

Basically:
* Create an aux route
* Create a datatable with virtual scroll, and create a routerLink to the aux route inside the datatable
* Click the link - it opens as expected
* Scroll down a few pages
* Click another link - note that the URL changes, but the component data never changes, and the ngOnInit hook never triggers
* Scroll in the table - the component now loads

Static text can still load, but none of the angular hooks are triggered.

**Expected behavior**
The aux route should load correctly, with the component initializing with the lifecycle hooks.

**Reproduction of the problem**
Rough repro @ https://stackblitz.com/edit/angular-eevppg

**What is the motivation / use case for changing the behavior?**
Ensuring the framework works as intended

**Please tell us about your environment:**
Windows 10, Webstorm, yarn, Angular CLI. Also reproducible in stackblitz ( see above )

* **Table version:** 0.8.x
@swimlane/ngx-datatable@^11.1.7

* **Angular version:** 2.0.x
"@angular/core@^5.1.3" (same for other packages)

* **Browser:**
Chrome (latest), Edge (latest)

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
Typescript 2.5

Fitkullin:
Bug Virtual scrollY (table 500 lines)
 <ngx-datatable-column name="Name" [flexGrow]="3" [minWidth]="200">
                    <ng-template let-value="value" let-row="row"  ngx-datatable-cell-template>
                        <a class="table-link"  [routerLink]="['/app', 'hr', 'employees', row.id]">{{row.name}}</a>
                    </ng-template>
</ngx-datatable-column>

after scroll down  routing stop working ,  ngOnInit hook never triggers

molcik:
Same here. Any workaround?

yusijs:
It seems to be related to zones. If you force the action to run within a zone, it works:

this.zone.run(() => this.navigate());

However, routerLinks etc wont work, and event emitters needs to be done within a function.

So if you did
(click)="select.emit(stuff)"
You now need to do
(click)="select(stuff)" in html, and
[...]

Issue #1216, opened 2018-01-02
Can't select a row when I scroll down with a mouse


gabrielnemeth:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
If you have datatable with scroll and selection enabled and if you scroll down, you can't select a row.

**Expected behavior**
You should be able to select a row even if you use scrolling.

**Reproduction of the problem**
[http://plnkr.co/edit/Uld4HKYkm0jiTJl5u7j1?p=preview](http://plnkr.co/edit/Uld4HKYkm0jiTJl5u7j1?p=preview)
- Scroll all the way down.
- Select one row by clicking on it
- The selection doesn't happen
- If you start scrolling, the selection appears

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

* **Table version:** 11.0.5 - 11.1.7

* **Angular version:** 5.1.1

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

* **Language:**  [all | TypeScript X.X | ES6/7 | ES5]

yusijs:
Think this is the same as my issue (https://github.com/swimlane/ngx-datatable/issues/1223) - something with zones, at least.

gabrielnemeth:
@yusijs using zone.run() doesn't fixed the issue for me. Maybe it's because I have issue with selection itself.

yusijs:
I've moved to row selections as well, and using zones works fine for me. 

However, it doesn't highlight the selection - perhaps it's the same for you? 

It seems related, nonetheless - to me it seems like the datatable detaches itself from angular when scrolling, and never reattaches itself. However that's just a wild guess.

Golosay:
Zone can help in this way:
```
public onSelect({ selected }): void {
    this.zone.run(() => {
      this.selected.splice(0, this.selected.length);
      this.selected.push(...selected);
    });
  }
```

sheiidan:
The zone issues might be related to this root cause:
https://github.com/swimlane/ngx-datatable/issues/1321#issuecomment-378011326

deeg:
If anyone is interested to use a version with scrolling running in zone, you can replace your package.json:

[...]

PR #1375, opened 2018-04-20
Do row rendering in angular zone

AsamK:
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When using virtual scrolling, after scrolling down, the event listeners for the newly rendered rows are created outside of the angular zone, which breaks automatic change detection from these event listeners.

Angular Material Tooltip stop working after virtual scrolling #1321
Virtual scroll causes issues with auxilary (named) routes #1223
Can't select a row when I scroll down with a mouse #1216
Dialog not working after scrolling #1348

**What is the new behavior?**
The scroll event is emitted from inside an angular zone so event emitters in the row templates are run the in the same zone and change detection works automatically.

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -412,7 +412,6 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
     }
 
     this.temp = temp;
-    this.cd.detectChanges();
   }
 
   /**
--- a/src/components/body/scroller.component.ts
+++ b/src/components/body/scroller.component.ts
@@ -46,9 +46,7 @@ export class ScrollerComponent implements OnInit, OnDestroy {
     if (this.scrollbarV || this.scrollbarH) {
       const renderer = this.renderer;
       this.parentElement = renderer.parentNode(renderer.parentNode(this.element));
-      this.ngZone.runOutsideAngular(() => {
-          this.parentElement.addEventListener('scroll', this.onScrolled.bind(this));
-        });
+      this.parentElement.addEventListener('scroll', this.onScrolled.bind(this));
     }
   }
 


bwegrzyn:
Would be nice to get this merged as this is currently breaking my entire application when scrolling in the table and clicking a row.

I've been forced to downgrade to 11.0.4 for the time being, but that has it's own set of bugs.

deeg:
@amcdnl, seems like this is causing a bunch of issues.

For me causing router links to not work once scrolling down.

I know you guys are super busy, just checking if this hasn't been merged due to no time, or if there is an issue with the fix itself. 

Would like to start a discussion of other possible solutions if this won't be merged due to it having performance implications.

marjan-georgiev:
@deeg we were trying to inspect any effect on the performance this might have. Running it in the zone would trigger angular's change detection, which is essentially what we want here, but we wanted to do more tests and understand how much this will affect performance, so we don't have to go back and revert it later on.

deeg:
@marjan-georgiev, thanks for the update. I'm currently doing the same and haven't seen any issue with making it run inside the zone.

I am not rendering millions of rows, but have tested it out with 1000+ rows rendered at once.

If anyone is interested to use a version with scrolling running in zone, you can replace your package.json:

`"@swimlane/ngx-datatable": "semanticresearch/ngx-datatable#scroll-in-zone",`

The only difference from release v13 is making scrolling happen inside zone:

https://github.com/semanticresearch/ngx-datatable/commits/scroll-in-zone

marjan-georgiev:
I think we're good to merge this. Thank you for the contribution!

PR #1478, opened 2018-07-03
datatable-row-wrapper: fix undefined reference access on group

tikri:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

We are using a caching service that only keeps the data of the viewport in memory. Before and after the viewport the array is empty. If scrolling up into the empty area of the data array, we get an exception ("unable to get property '0' of undefined").

**What is the new behavior?**
No exception is thrown when scrolling up the table.

- [ ] Yes
- [x] No

If this PR contains a breaking change, please describe the impact and migration path for existing applications: -

**Other information**:


Hypercubed:
Thank you for the PR.... this will be fixed in this PR: #1853

PR #1492, opened 2018-07-20
Add canShowDataTable getter/setter to show/hide datatable element.

Samuel-Ayvazyan:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The datatable container always visible

**What is the new behavior?**
Using canShowDataTable getter/setter we can show/hide datatable container

- [ ] Yes
- [x] No

**Other information**:
It is very convenient to have some general functionality on the application to show/hide datatable container depends on some conditions without hiding full component and losing data or space which it occupies.

### DIFF
--- a/src/components/datatable.component.ts
+++ b/src/components/datatable.component.ts
@@ -19,12 +19,12 @@ import { DatatableRowDetailDirective } from './row-detail';
 import { DatatableFooterDirective } from './footer';
 import { DataTableHeaderComponent } from './header';
 import { MouseEvent } from '../events';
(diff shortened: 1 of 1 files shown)


eppsilon:
Why not just put a `*ngIf` on the `<ngx-datatable>` in the code using the table component?

Samuel-Ayvazyan:
> Why not just put a `*ngIf` on the `<ngx-datatable>` in the code using the table component?

It is a much better approach to use getter/setter programmatically instead of simple true/false in *ngIf

PR #1493, opened 2018-07-20
Fix issue with accessing data of async rows.

Samuel-Ayvazyan:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, the template of body component is trying to access to `group[i]` without checking group and when data is coming async we are getting errors like this:
ERROR TypeError: Cannot read property '0' of undefined

**What is the new behavior?**
Check for 'group' variable and only after try to access the index of it.

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -46,10 +46,10 @@ import { MouseEvent } from '../../events';
           [rowDetail]="rowDetail"
           [groupHeader]="groupHeader"
           [offsetX]="offsetX"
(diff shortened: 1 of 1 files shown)


tikri:
nearly same approach as #1478 👍

Hypercubed:
Thank you for the PR.... this will be fixed in this PR: https://github.com/swimlane/ngx-datatable/pull/1853

PR #1495, opened 2018-07-24
Fix flex column resize

Toktik:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When the column mode is flex, the resize on columns does not work

**What is the new behavior?**
Column resize fixed in case of flex column mode

- [ ] Yes
- [x] No

**Other information**:


EduardsE:
Tried forking repo, committing these changes, using the repo as dependency in my project, but it still wouldn't work.

Issue #1158, opened 2017-11-28
HTML Content is parsed in the body-cell


fsc93:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
When I create a datatable with a column like: columns = [{ prop: 'foo' }];
and a row like: rows = [{ foo: '\<strong>some text\</strong>'}];
The text of the foo property is displayed as: <strong>some text</strong>
instead of: \<strong>some text\</strong> as it is bound to the innerHTML of the span inside the
BodyCellComponent.

**Expected behavior**
The property value should be set as the spans textContent so the HTML won't get parsed.

**Reproduction of the problem**
create a datatable like this: 
<ngx-datatable width="100%" [rows]="rows" [columns]="columns"></ngx-datatable>

with public columns = [{ prop: 'foo' }];
and public rows = [{ foo: '\<strong>some text\</strong>' }];

**What is the motivation / use case for changing the behavior?**
Users can edit the data displayed in the table so html content should not be parsed.

**Please tell us about your environment:**

* **Table version:** 9.3.1

* **Angular version:** 4.2.6

* **Browser:** Chrome 62

* **Language:** TypeScript 2.3.4

555ea:
This makes column cells vulnerable to XSS attacks. Is this issue going to be fixed ? @amcdnl 

The tooltip is paste like a sanitizedValue, why not sanitize [innerHTML] ? https://github.com/swimlane/ngx-datatable/blob/master/src/components/body/body-cell.component.ts#L28

There is workaround of passing a template or escaping pipe, but both of them create a performance overhead.

ayazhussein:
any updates on this?

atakchidi:
What's the problem here? If you need to show html as a text add column template for that column. This use case is rare. IMHO innerHtml is safer, and since you always can pass a template with whatever you wan't theres no issue

jeremyben:
When you use the [innerHtml] directive, as ngx-datatable does, Angular already sanitizes the html content beforehand, so no vulnerability to XSS attacks.

See:
https://angular.io/guide/security
https://angular.io/api/platform-browser/DomSanitizer

defhead:
Hello 

[...]

PR #1536, opened 2018-08-31
Don't parse HTML Content in body cell #1158

BrummbQ:
This straight up addresses issue #1158 For discussion and reasoning see there


jeremyben:
Like I said in the issue :
When you use the [innerHtml] directive, as ngx-datatable does, Angular already sanitizes the html content beforehand, so no vulnerability to XSS attacks.

See:
https://angular.io/guide/security
https://angular.io/api/platform-browser/DomSanitizer

I think this will limit a ngx-datatable feature (direct output of html in a cell) without adding security as expected.

Issue #1196, opened 2017-12-14
Row Grouping and Row Details not working together


pssubashps:
**I'm submitting a ...**  (check one with "x")
```
[x ] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**

the row details open by default with out marking for expand
**Expected behavior**

It should open only when click on the icon
**Reproduction of the problem**

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

* **Table version:** 0.8.x

10.4.0
* **Angular version:** 2.0.x
4.3.0

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

Chrome Version 63.0.3239.84 (Official Build) (64-bit)
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

pssubashps:
```
> <ngx-datatable-row-detail [rowHeight]="120">
>    <ng-template let-row="row" #rowdetails let-expanded="expanded" ngx-datatable-row-detail-template>
>            <ng-container  *ngTemplateOutlet="rowDetailsTemplate"
>            ></ng-container>
>           </ng-template>
>       </ngx-datatable-row-detail>
>       <!-- Group Header Template -->
>     <ngx-datatable-group-header [rowHeight]="50" #myGroupHeader>
>       <ng-template let-group="group" let-expanded="expanded" ngx-datatable-group-header-template>
>         <div style="padding-left:5px;">
>           <a
>             [class.datatable-icon-right]="!expanded"
>             [class.datatable-icon-down]="expanded"
>             title="Expand/Collapse Group"
>             (click)="toggleExpandGroup(group)">
>             <b>Status: {{group.value[0].isGlobal}}</b>
>           </a>                          
>         </div>
>       </ng-template>
>     </ngx-datatable-group-header>
```

      

here is the issue.
both are getting the bind with the same property " expanded.". so that is it showing
any suggestions ?

Thanks,
Subash

sapottier:
A temporary workaround

`<ngx-datatable-row-detail #rowDetail [rowHeight]="'auto'">
            <ng-template let-row="row" let-rowIndex="rowIndex" ngx-datatable-row-detail-template>
[...]

Issue #1108, opened 2017-11-07
Row grouping with row detail issue


kresogalic8:
Hey guys, I've used your library and I wanted to achieve to group one row and also that row can toggle row detail.

In that case, I will get row grouping but row detail will be immediately opened. Is there any solution for that? I've followed this http://swimlane.github.io/ngx-datatable/#row-grouping to achieve row grouping.

pssubashps:
I also facing the same issue

zagros:
same issue (+1)

Logic-Bits:
same here

Ggleny:
same issue (+1)

PR #1569, opened 2018-10-08
Nest the row-wrapper so that we can have a row-detail tied to each ro…

marcushill:
…w of a group

- [X] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

#1196 #1108 

**What is the new behavior?**
Each row of a group gets its own datatable-row-wrapper. What this effectively means is that for groups specifically, you have the following HTML:
```html
<datatable-row-wrapper>
   <datatable-row-wrapper>
       <datatable-body-row></datatable-body-row>
   </datatable-row-wrapper>
</datatable-row-wrapper
```

- [ ] Yes
- [x] No

To the best of my ability to check, this should not break any existing functionality outside of someone's table styles that use child selectors rather than descendant selectors to style the datatable-body-row. I have run the demo page with this change and they don't render any differently. 

**Other information**:
I have also backported this to 11.x as my project is still on Angular 5 and we had (currently undiagnosed) issues running the latest in production.


med925:
Any news here?

Issue #1049, opened 2017-10-11
Making a change to a row in a collapsed group causes the group to expand


rwhere:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**

In regards to row grouping: if you collapse a group, and make a change to any cell in the collapsed group, it will expand the group

**Expected behavior**

I expect the group to remain collapsed

**Reproduction of the problem**

1.) go to http://swimlane.github.io/ngx-datatable/#row-grouping
2.) collapse group Age: 22
3.) make a change to any cell in the comment column for the groups that remain expanded
4.) click outside the cell for the changes to take effect
5.) the Age: 22 group will expand

**What is the motivation / use case for changing the behavior?**

Not everyone is building a static table. I update the values of a certain column every few milliseconds. So as soon as a I collapse a group, it expands immediately after.

**Please tell us about your environment:**

Operating System: macOs Sierra (10.12.6)
IDE: Visual Studio Code
Package Manager: NPM (3.10.10)

* **Table version:** 0.8.x

ngx-datatable 10.2.3

* **Angular version:** 2.0.x

4.0.0

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

all

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

typescript: 2.5.3

rwhere:
@CunhaGus

am74po:
Hi.
I have the same issue...
Just upgraded to version 14.0.0 , but I still can see the clear() command when setting the _rows.
Please advise.
Thanks.

4nd3r5:
A little late for a comment but just a reccomendation: I kept the changeable values for the rows in a different datatype, with the rowid as a lookup.
For instance I had a map string -> boolean for whether a value was checked or not.
That way, I did not change the values in the row and thus no clear command was fired.

PR #1583, opened 2018-10-23
#1049 Preserve group expansion state when updating rows

peterhorsley:
Use row key as key for rowExpansions map to ensure that group expansion state is maintained across row data updates.

Fixes issue #1049 Making a change to a row in a collapsed group causes the group to expand

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue is reproducible on demo page as per instructions here:
https://github.com/swimlane/ngx-datatable/issues/1049

**What is the new behavior?**
Data updates no longer cause groups to be collapsed/expanded.  Newly added groups will be set to the defined default expansion state as per table config.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/components/body/body.component.ts
+++ b/src/components/body/body.component.ts
@@ -146,7 +146,6 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
 
   @Input() set rows(val: any[]) {
     this._rows = val;
-    this.rowExpansions.clear();
     this.recalcLayout();
   }
 
@@ -452,7 +451,7 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
    */
   getRowAndDetailHeight(row: any): number {
     let rowHeight = this.getRowHeight(row);
-    const expanded = this.rowExpansions.get(row);
+    const expanded = this.rowExpansions.get(row.key);
 
     // Adding detail row height if its expanded.
     if (expanded === 1) {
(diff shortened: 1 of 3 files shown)


src/components/body/body.component.ts
```diff
@@ -146,7 +146,6 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
 
   @Input() set rows(val: any[]) {
     this._rows = val;
-    this.rowExpansions.clear();
```
markwaddle: 🙏

src/components/body/body.component.ts
```diff
@@ -452,7 +451,7 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
    */
   getRowAndDetailHeight(row: any): number {
     let rowHeight = this.getRowHeight(row);
-    const expanded = this.rowExpansions.get(row);
+    const expanded = this.rowExpansions.get(row.key);
```
markwaddle: adding a dependency on an arbitrarily named field seems fragile and non-obvious. what if you used the rowIdentifier function to get the key?

peterhorsley: hey mark, thanks for taking a look at this PR.  getRowAndDetailHeight is called by getGroupHeight which is actually passing not a regular row, but a 'grouped row' which is created by groupArrayBy in datatable.component.cs.  That function creates a may with key/value pairs, where 'key' has the value of the unique set of values defined by the 'groupByRows' property, and value is the list of rows that match that key.  So in this case, key always exists and it intentional to query the rowExpansions by key as this is the way they are grouped.  Hope that makes sense!  I did just notice in this commit I accidentally included an fdescribe in the tests, will remove that now.

peterhorsley: @markwaddle hi mark do you need any further clarification, would be great to get this PR merged if you are happy with it.

pudding157:
Bump, is this ever getting released?

Issue #1117, opened 2017-11-13
New Feature Request:  Ability to "select" a row programmatically.


allen0977:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[ x ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**

Unable to programmatically "Select" a row.   In order to perfom the "Select" operation, the user must click the row.

**Expected behavior**

Within the TS file, I'd like to perform a "Select" operation on a specific row, and have it "Open" the details of that row.   I would like to be able to SAVE the selected row parameter, perform an update of the underlying this.rows, and then reselect the previously selected rows.  

**Reproduction of the problem**

**What is the motivation / use case for changing the behavior?**

In my application, the rows display summary data for multiple months worth of information.  Upon selecting a row, it opens a detailed datatable, that shows projects and values for expanded for the same months.   Within the details table, users can interact with the data and update values, which commit to the database.   A refresh of the original data table data will force the detail view to close.   I would like to keep it open (or at the very least be able to "reselect" the record that was open in the first place).

**Please tell us about your environment:**

Mac, Sublime/Ultra Edit, npm, node, etc.

* **Table version:** 0.8.x

11.0.2 is the current version I'm using, this is a new feature request not found in older versions.

* **Angular version:** 2.0.x

5.0.1

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

All

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
all

nhducseuit:
The DatatableComponent object should allow you to do that.

Assume that you already have object `row` that you want to select it from ts file, and by using `@ViewChild` query you able to obtain an instance of the table named `datatable`

In order to select the row from ts, do this:
```
this.datatable.selected.push(row);
```

In order to navigate to a page with specific selected row info, just use Angular router feature, like:

```
this.router.navigate(['app/detailpage', row.id]);
```

ashishpok:
@nhducseuit solution seem to need at least another keyboard action - meaning it sets the selection only after I hit key up/down etc. Is there is way to refresh? I tried pushing rows back into the table with rows = [...rows] and that didnt seem to work either.

nhducseuit:
[...]

Issue #872, opened 2017-07-21
Scroll to newly inserted (selected) row (programmatic row focus)


ronnyek:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[x ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
While there doesn't appear to be api's to insert a specific row, its easily accomplished by inserting rows into the dataset. What I'd ultimately like to be able to do is scroll to newly inserted rows... (and that could be as simple as just creating the row data, inserting, setting that as the selected item. (so no need to build functionality to specifically scroll to an inserted or selected record).

**Expected behavior**
What I'd liked to be able to do is provide it a row, or index or something, and be able to scroll to that item. When the row item is displayed (virtual scrolling), I could technically get a reference to `datatable-body-row.datatable-body-row.active`, but naturally that dissapears as soon as that row is scrolled out of view.

**What is the motivation / use case for changing the behavior?**
Current API lets you programatically select a row, but if that row happens to scrolled off the screen whilst selected element is set, its not clear there is a selection

**Please tell us about your environment:**

* **Table version:** 0.8.x
Currently 9.0.0

* **Angular version:** 2.0.x
4.0.0

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
all browsers should be affected as this appears to be new functionality

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
typescript 2.3.2

ronnyek:
I just wrote a test... aside from from inserting etc, if you have a table that say loads up the company.json, and set that to virtual scroll with a height of say 300, set the selected record as the last loaded item... it does indeed select that record and show that as the selected record, but you may not know it because you'd have to scroll down to the end of the list to see it selected.

amcdnl:
A external API to jump to a row index might be the desired result here. Shouldn't be too hard to implement if you wanna PR that.

szpetny:
Hi, what is a status of this? I would also be interested in something which lets jump to a selected row.

szpetny:
So it looks like it is possible to do it like this: 
`tableEl.bodyComponent.updateOffsetY(row.index/tableEl.bodyComponent.pageSize)`

[...]

Issue #851, opened 2017-07-07
Datatable component blocks arrow events


szanella:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[x] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
I'm using the datatable component to display a set of rows, and the (activate) output to be notified of row clicks: when a row is clicked, I open a side panel that displays additional data, and I implemented it so that it listens to keydown events to move between the table rows when it's open. This was not working, though, as the table, being focused, would eat up all the keydown events to "do its thing" and would stop their propagation.

**Expected behavior**
I would like this propagation stopping to be configurable.
More specifically, I think the table should only react to keydown events (and therefore block them) if the developer wants it to.

**Reproduction of the problem**
A bit too complex to set up.

**What is the motivation / use case for changing the behavior?**
It's not possible to use keydown events on arrow keys.

**Please tell us about your environment: Max Os X El Capitan, Angular CLI 1.0.0

* **Table version:** 9.0.3

* **Angular version:** 4.2.5

* **Browser:** Chrome 59.0.3071.115

* **Language:** TypeScript 2/2

dannyvenier:
can't you use the activate output event to indicate to the side panel which row is in focus?

szanella:
You definitely have a point, but the problem come as soon as the table loses focus.
Take this course of events as an example:
- I click on a table row, which opens my side panel over the rest of the window (the table still has focus, though)
- I use the arrow keys to navigate the rows, which triggers the activate event and allows me to navigate it through events fired by the activate output
- I use the buttons on the side panel, which of course makes the table lose the focus, which now goes on the panel component
- having lost the focus, the table will no longer be able to receive keydown events

My solution, in fact, was implementing a directive that, when the panel opens, moves the focus on it.

dannyvenier:
Well, it's hard to picture your application without seeing it, but unless the table is a child of the side panel, they are not in the same event chain so the key events will never make it to the table if the focus is on the panel.  I can't picture the use case but if scrolling the table while focused on the panel is something you need to do, I would have key listeners on the panel and invoke controls on the table from the side panel handlers (either through a shared service or through an Observer which emits between peers).  Or you can implement up and down buttons on the panel to invoke the iteration through the table.  If the panel covers the table you are navigating blind but again, I can't picture the use case.

[...]

PR #1639, opened 2019-01-30
Provide access to Selector via ViewChild

msnider:
- [ ] Bugfix
- [X] Feature
- [-] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

Currently there is no way to get access to the DataTableSelectionComponent. This prohibits users from creating custom select row logic. There are several open issues that this could facilitate a workaround for including #1117, #851, #872. 

**What is the new behavior?**
Only added a @ViewChild reference in the DataTableBodyComponent to the DataTableSelectionComponent. This allows users to select rows via: `table.bodyComponent.selector.selectRow(event, idx, row)`. I'm currently using this with the activate output to navigate the selected row using the keyboard and it works great including across virtual pages.

- [ ] Yes
- [X] No

If this PR contains a breaking change, please describe the impact and migration path for existing applications: None.

**Other information**:
Probably not a long term solution given this is supporting an undocumented API.


carter-dewey:
Would love for this to be added into the data-table

bot101:
This feature needs to be added. Pretty important. Any other ways to select rows from logic pending when this it finally merged?

PR #1660, opened 2019-02-27
Remove usage of @Input and @ContentChild on the same field, for Ivy

alxhub:
ngx-datatable used to use @Input and @ContentChild on the same field. This
pattern is only allowed by Angular on accident, and is rejected by Ivy's
stricter validation of application & library code.

This commit replaces the pattern with separate fields for the @Input and
@ContentChild query, and a getter which selects whichever one is set. The
input is allowed to override the query.

I verified this passes all the unit tests. The e2e tests seem broken (they
fail without this change on my machine).

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`@Input` and `@ContentChild` are used on the same field. This is accepted by Angular today, but the fact that it works is generally an accident. Angular Ivy doesn't support a query and an input both on the same field.

**What is the new behavior?**

Each of the 6 fields which have both annotations is split into two fields, one for the `@Input` and one for the `@ContentChild`. A getter is introduced which selects whichever of the two is set, with a preference given to the `@Input`.

- [ ] Yes
- [x] No

**Other information**:


marjan-georgiev:
@alxhub thank you for the contribution!

PixelShard:
What about to release new version with this fix for IVY compiler?

Issue #1686, opened 2019-05-13
column resizing issue in force columns mode (no horizontal scroll) when container get smaller


barouchmaximedenoun:
I find an issue in force mode column recalculation and I can also propose a fix for this issue. I will submit soon (tomorrow) a Pull Request to be reviewed.

The following example is to demonstrate the issue and is very little coding:
https://stackblitz.com/edit/angular-2e7nxq

In force columns mode with no horizontal scroll, when the table container is getting bigger (when I am hiding the slider clicking Show/Hide Slider button), I am sending a window resize event (in method triggerTableResize in app.component.ts) and the table resize its column perfectly accordingly to the new container width. But if I shrink the container (expanding the slider clicking a second time on Show/Hide Slider button) the table resize its column with a little miscalculation (the ellipsis button in the table is cut).

The problem is in ngx-datatable forceFillColumnWidths in utils.Math.ts,

When the remainingWidth is positiive we will go several time in 
do {
} while (remainingWidth > 0 && columnsToResize.length !== 0);

but when the container shrink and the remainingWidth is negative we just go once.

i will submit a Pull Request tomorrow with a fix.

resize the window which adapt the size of the column

**I'm submitting a ...**  (check one with "x")
```
[x ] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, post on Stackoverflow or Gitter
```

**Current behavior**
https://stackblitz.com/edit/angular-2e7nxq
clicking twice Show/Hide Slider button make the last column cut.
expand a bit the window the last column is not cut anymore.

**Expected behavior**
The last column is not cut.

**Reproduction of the problem**
https://stackblitz.com/edit/angular-2e7nxq
click twice on Show/Hide Slider button.
look at last column with ellipsis button which is cut

**What is the motivation / use case for changing the behavior?**

when the table container shrink because of a slider or going from full screen to normal screen  clciking on a button the last column will always be cut.

**Please tell us about your environment:**

reproductible on stackblitz.com also in my environment, Windows 10 Pro, Webstorm, npm
* **Table version:** 

14.0.0
* **Angular version:** 

v7

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

Chrome at least
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
[...]

PR #1687, opened 2019-05-14
fix issue 1686: column resizing issue in force columns mode (no horiz…

barouchmaximedenoun:
fix column resizing issue in force columns mode (no horizontal scroll) when container get smaller #1686

https://github.com/swimlane/ngx-datatable/issues/1686

- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

https://github.com/swimlane/ngx-datatable/issues/1686

**What is the new behavior?**

https://github.com/swimlane/ngx-datatable/issues/1686

- [ ] Yes
- [ x] No

**Other information**:

### DIFF
--- a/src/utils/math.ts
+++ b/src/utils/math.ts
@@ -112,6 +112,7 @@ export function forceFillColumnWidths(
   let contentWidth = getContentWidth(allColumns, defaultColWidth);
   let remainingWidth = expectedWidth - contentWidth;
   const columnsProcessed: any[] = [];
+  const remainingWidthLimit = 1; // when to stop
 
   // This loop takes care of the
   do {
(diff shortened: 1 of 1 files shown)


marjan-georgiev:
Thank you! Will be released in 15.1.0

src/utils/math.ts
```diff
@@ -141,7 +142,7 @@ export function forceFillColumnWidths(
     contentWidth = getContentWidth(allColumns);
     remainingWidth = expectedWidth - contentWidth;
     removeProcessedColumns(columnsToResize, columnsProcessed);
-  } while (remainingWidth > 0 && columnsToResize.length !== 0);
+  } while (Math.abs(remainingWidth) > remainingWidthLimit && columnsToResize.length !== 0);
```
vixriihi: Why the Math.abs? In our case expectedWidth is negative (seems like it's counted when column mode is forced & scrollbarV this is counted as innerWidth - scrollbarHelper.width) and that will result to an infinite loop.

ScottSpittle: Yeah this caused an infinite loop with my usage so I've reverted it locally to ensure I can still use this library

clegeard:
How is this fixed?
I'm still having the same problem with version 20.0.0
I am using the same kind of design, a slider on the left of a table appearing on a particular event.
When this happens the columns are not resized and thus the table neither.
If you run the example provided in the original issue: https://stackblitz.com/edit/angular-2e7nxq
The table is not resized, there is a slider appearing on it.
Could you please revisit?

PR #1721, opened 2019-07-13
Pass extra columns as input to table + columns optional initial order

duredhel13:
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

**What is the new behavior?**

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/docs/api/column/inputs.md
+++ b/docs/api/column/inputs.md
@@ -1,70 +1,96 @@
 # Table Column Options
 
 ### `name`: `string`
+
 Column label. If none specified, it will use the prop value and decamelize it.
 
 ### `prop`: `string`
+
 The property to bind the row values to. If `undefined`, it will camelcase the name value.
 
 ### `flexGrow`: `number`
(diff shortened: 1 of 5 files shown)


demo/basic/reusability/reusability.component.ts
```diff
@@ -0,0 +1,50 @@
...
+      <reusable-table [columns]="columns" [rows]="rows">
+        <ngx-datatable-column name="Gender"></ngx-datatable-column>
+        <ngx-datatable-column name="Company"></ngx-datatable-column>
+      </reusable-table>
+    </div>
+  `
+})
+export class ReusabilityComponent {
+  rows = [];
+
```
syanpriyajot: Is it really reusable when you are hard-coding the column values at component level.
The reason i said this because i tried using the same thing but the passing the data via @Input property and it throws error

duredhel13: The columns passed via @input need to be of type `QueryList<DataTableColumnDirective>`, I assume it will fail any other way.

PR #1770, opened 2019-10-28
Guard setStylesByGroup if columns not yet set

atscott:
If you call calcStylesByGroup before the `columns` Input is set, `this._columnGroupWidths` will be undefined and the component will throw an error. Specifically the `offsetX` input unconditionally calls `calcStylesByGroup` and will fail in the upcoming version of Angular because the input order changes in Ivy. Inputs should avoid depending on other inputs as much as possible or guard against unexpected input orders.

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Component throws an error in Angular Ivy if offsetX is bound before columns


atscott:
@marjan-georgiev Could you take a look at this? If this issue is not addressed, the library will not work with Angular v9

PR #1815, opened 2020-01-27
Fix column width recalculation

ErikStelzerKomsa:
- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Problem: When reducing the size of the browser window, the rightmost column(s) are sometime not visible. Resizing the window larger and smaller causes the columns to be recalculated. The user may have to repeat this several times until the columns are correct.

Probably cause: When calculating the column widths, it is always based on the total width of all columns:
additionWidthPerColumn = remainingWidth / columnsToResize.length;

Columns with minWidth or maxWidth are set to their defined value:
column.width = column.minWidth;
column.width = column.maxWidth;

Only columns without minWidth or maxWidth are recalculated:
newSize = (column.width || defaultColWidth) + additionWidthPerColumn;
…
column.width = newSize;

**What is the new behavior?**

To fix this, the total width to be portioned out should only be used for columns can be resized (i.e. have a recalculated newSize).

- [ ] Yes
- [x ] No

**Other information**:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ coverage/
 .idea
 yarn.lock
 /.vscode
+/.vs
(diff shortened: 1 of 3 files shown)


Iuriy-Budnikov:
This pr doesn't fix blank issue after resize

PR #653, opened 2017-03-31
Allow for dynamic Detail Row heights

yobryon:
**What kind of change does this PR introduce?** (check one with "x")
- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

**What is the current behavior?** (You can also link to an open issue here)
detailRowHeight is declared in the ngx-datatable-row-detail component as a number and is applied globally to all row detail panes that are shown in the table. 

**What is the new behavior?**
Allows for existing use case (no breaking), and introduces ability to specify rowHeight as a callback function which is injected with information from the row wrapper iteration (specifically, row and index). In this way, we can make a decision about the height of a given detail row based on logic that may reference the data row (or not -- it's up to the dev). 

Example:
HTML
`            <ngx-datatable-row-detail [rowHeight]="getRowHeight">
              <ng-template ngx-datatable-row-detail-template let-row="row">
                <md-list dense > 
                  <md-list-item *ngFor="let item of row.items">{{item}}</md-list-item>
                </md-list>
              </ng-template>
            </ngx-datatable-row-detail>`

TS
`  getRowHeight(row:any,index:number):number{
    return row && row.items? row.items.length*50 : 0;
  }`

**Does this PR introduce a breaking change?** (check one with "x")
- [ ] Yes
- [x] No

If this PR contains a breaking change, please describe the impact and migration path for existing applications: ...

**Other information**:

PR #1893, opened 2020-06-05
Fixed invalid argument when rowHeight is a function

ANovokmet:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When rowHeight is passed as a function to ngx-datatable-row-detail, it is called with the parameter row undefined, which makes it impossible to dynamically calculate height of row detail based on row data.

This is happening because when rows are not grouped, method getDetailRowHeight of datatable-body is called incorrectly.

**What is the new behavior?**

When rowHeight is a function it is called with the current row as the parameter.

- [ ] Yes
- [x] No

**Other information**:


ANovokmet:
https://github.com/swimlane/ngx-datatable/pull/653#issuecomment-385871156 is caused by this problem.

PR #2064, opened 2022-01-24
Add all components/directives to module exports

yusijs:

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [X] Other... Please describe: See under "Other information"

All components and directives are not exported

**What is the new behavior?**
All components and directives are exported

- [ ] Yes
- [X] No

**Other information**:

Functionally this makes no difference for consumers of the library. However, we are building some features on top of this table that are very specific to our usecase. In other words, getting our changes into this project is not feasible. We're unable to build our features on top of this table without cloning the repo into our own code. Obviously this is not a good solution in the long-term. By exporting the components, we're able to reuse everything from this library and only override the ones we need. Without exporting them the components are unknown to angular.


Piyush132000:
Hii , Is anyboday working on it. if you don't mind i want to fix this one. As I see 9 of 2 completed issue headline.

PR #2133, opened 2023-01-02
Allow 'auto' to be passed as a height

Whatshiywl:

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

The [headerHeight](https://swimlane.gitbook.io/ngx-datatable/api/table/inputs#headerheight) and [footerHeight](https://swimlane.gitbook.io/ngx-datatable/api/table/inputs#headerheight) inputs of `ngx-datatable` accept the 'auto' value as well as numbers, going as far as [verifying if their values match that specific string](https://github.com/swimlane/ngx-datatable/blob/master/projects/swimlane/ngx-datatable/src/lib/components/datatable.component.ts#L467). However, their current type signature allows for numbers only. This type conflict pops up when using `strictTemplates` as an [Angular Compiler Option](https://angular.io/guide/angular-compiler-options#stricttemplates).

**What is the new behavior?**
The two aforementioned inputs will now allow the 'auto' value to be passed without incurring in build errors when template type checking is enabled.

- [ ] Yes
- [X] No

### DIFF
--- a/projects/swimlane/ngx-datatable/src/lib/components/datatable.component.ts
+++ b/projects/swimlane/ngx-datatable/src/lib/components/datatable.component.ts
@@ -189,13 +189,13 @@ export class DatatableComponent implements OnInit, DoCheck, AfterViewInit {
    * The minimum header height in pixels.
    * Pass a falsey for no header
    */
-  @Input() headerHeight: number = 30;
+  @Input() headerHeight: number | 'auto' = 30;
 
   /**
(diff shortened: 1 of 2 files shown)


achimha (changes requested):
if headerHeight/footerHeight is 'auto', the calculation with `height = height - (this.footerHeight as number)` will not work.

PR #2138, opened 2023-02-02
Fix memory leaks

Arooba-git:

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Significant memory leakage

**What is the new behavior?**
Significant reduction in memory leakage

- [ ] Yes
- [x] No
-----------------------------------
Hello 👋 
As part of our project, we are using Facebook's new Memlab tool to detect memory leaks in SPA applications and their libraries. While running the tool and analyzing the code of ngx-datatable, we saw that your project does a very good job of ensuring that all async operations are cancelled when components destroy. However, as per Memlab execution results, we found some dangling animation frame requests, timeout calls and QueryLists, that were causing the memory to leak (screenshots below).

[before]
Screen Shot 2023-02-03 at 7 49 08 AM

Screen Shot 2023-02-03 at 8 32 11 AM

Hence we added the fix by clearing the frame requests, timeout calls and querylists in destructor, and you can see the heap size and # of leaks reducing noticeably:
 <br />
Screen Shot 2023-02-03 at 6 11 11 AM

Note 1. Host listener is used to ensure that ngOnDestroy is called not only when the class destroy, but also when the page refreshes, tab or browser closes or the user navigates away from the page[1]

Note 2. Static property is used to avoid losing the context of the correct ‘this’ in case any nesting changes are made to the enclosing function in future.

We executed the test suite before and after the changes, where it executed 126 of 129 (skipped 3, TOTAL: 126 SUCCESS) in both cases.

You can analyze this and other potential leak sources, if you like, by running [Memlab](https://facebook.github.io/memlab/) with a scenario file covering the maximum # of use cases.

Following is a sample of the scenario file we used (it needs to be a .js file but attaching here in .txt form):
[ngx-database-scenario-memlab.txt](https://github.com/swimlane/ngx-datatable/files/10574350/ngx-database-scenario-memlab.txt)

Note that some other reported leaks (in Memlab) originated from Angular's internal objects, hence were ignored.

Reference.
[1] https://wesleygrimes.com/angular/2019/03/29/making-upgrades-to-angular-ngondestroy.html

### DIFF
--- a/projects/swimlane/ngx-datatable/src/lib/components/body/body.component.ts
+++ b/projects/swimlane/ngx-datatable/src/lib/components/body/body.component.ts
@@ -122,6 +122,7 @@ import { translateXY } from '../../utils/translate';
   }
 })
 export class DataTableBodyComponent implements OnInit, OnDestroy {
+  static _timeOutID: any;
   @Input() scrollbarV: boolean;
   @Input() scrollbarH: boolean;
   @Input() loadingIndicator: boolean;
@@ -327,6 +328,7 @@ export class DataTableBodyComponent implements OnInit, OnDestroy {
     if (this.rowDetail || this.groupHeader) {
       this.listener.unsubscribe();
     }
+    clearTimeout(DataTableBodyComponent._timeOutID);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


Arooba-git:
bump @haeminn

alybal (approved):

PR #2236, opened 2025-05-06
(release): 21.0.0

steveblue:

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

**What is the new behavior?**

This PR upticks the package for v21.

## 21.0.0

- Fix: Add missing package files from deployment
- Breaking: support for angular 17-19

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Run `yarn build` to build the project. The build artifacts will be stored in the
 - Examine log to determine next version (X.Y.Z)
 - Run `git checkout -b release/X.Y.Z`
 - Update version in `projects/swimlane/ngx-datatable/package.json`.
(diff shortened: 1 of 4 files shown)


jogaj (approved):

bal3no:
Hi, I've noticed a new (and for me unwanted) behaviour on the sort event. If you select a column, clicking near it's name and without sorting it, then you can invoke the sort event by pressing the enter key. In my own case, i've put an input box under each column, in the header, to be able to filter it. My filter triggers after the keyboard enter key and this, produces an unwanted sorting event on that column.
Tested also in the table's demo online page and produces the same behaviour.

